PR #3, opened 2020-10-27
Fix crash when ssl parameter of api is used

felixfontein:
##### SUMMARY
The problem is that the function parameter is called `ssl`, which hides the `ssl` module which is also used in that function. This results in `error: 'bool' object has no attribute 'create_default_context'`.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
api


NikolayDachev (approved):
I just notice that as well .. thanks !

felixfontein:
@NikolayDachev feel free to press the Squash+Merge button :)

felixfontein:
@NikolayDachev thanks for reviewing and merging! Also congrats on your first merge in this repo ;)

PR #21, opened 2021-03-23
filter out datetime from facts config

### LINKED ISSUE #20: facts subset config is allways "changed"
##### SUMMARY

Every time 'community.routeros.facts' subset 'config' returns data it's status is 'not changed' but it is false.
Returned data are different on every run of 'community.routeros.facts' even configuration on device is not changed:
1. there is date at the beggining of the file
2. with "/export verbose" there are exported unconfigured services too and for example unconfigured '/interface ovpn-server server' contains random generated MAC address (tested on RouterBOARD 750G r3)

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

'community.routeros.facts' subset 'config'

##### ANSIBLE VERSION

```paste below

```

##### CONFIGURATION

```paste below
# ansible --version
ansible 2.10.6
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.8/dist-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 3.8.5 (default, Jul 28 2020, 12:59:40) [GCC 9.3.0]
```

##### OS / ENVIRONMENT

Ubuntu 20.04
RouterBOARD 750G r3
RouterOS 6.40.3

##### STEPS TO REPRODUCE

setup your inventory to an Mikrotik device (tested on RouterBOARD 750G r3)

```yaml
---
- name: Mikrotik config1
  community.routeros.facts:
    gather_subset: config

- set_fact: cfg1="{{ ansible_net_config }}" 
- set_fact: cfg3="{{ ansible_net_config }}" 

- name: Mikrotik config2
  community.routeros.facts:
    gather_subset: config

- set_fact: cfg2="{{ ansible_net_config }}" 

- ansible.builtin.debug: 
    msg: "cfg1 is different from cfg2"
  when: cfg1 != cfg2

- ansible.builtin.debug: 
    msg: "cfg1 is not different from cfg1"
  when: cfg1 == cfg1

- ansible.builtin.debug: 
    msg: "cfg1 is not different from cfg3"
  when: cfg1 == cfg3
```

##### EXPECTED RESULTS

cfg1 should be the same cfg2 because device configuration did not changed

##### ACTUAL RESULTS

cfg1 and cfg2 are different.
it makes very difficult to version those cfg files

qaxi:
##### SUMMARY

Every time 'community.routeros.facts' subset 'config' returns data it's status is 'not changed' in persptective of device it is OK, but ...
Returned data are different on every run of 'community.routeros.facts' even configuration on device is not changed

Fixes #20 part 1.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

'community.routeros.facts' subset 'config'

##### ADDITIONAL INFORMATION

```BEFORE
# mar/22/2021 18:23:30 by RouterOS 6.40.3
# software id = 2TZZ-SFZA
#
# model = RouterBOARD 750G r3
# serial number = 6F3806FD9931
/interface ethernet
set [ find default-name=ether1 ] advertise=\
    10M-half,10M-full,100M-half,100M-full,1000M-half,1000M-full arp=enabled \
```
```AFTER
# RouterOS 6.40.3
# software id = 2TZZ-SFZA
#
# model = RouterBOARD 750G r3
# serial number = 6F3806FD9931
/interface ethernet
set [ find default-name=ether1 ] advertise=\
    10M-half,10M-full,100M-half,100M-full,1000M-half,1000M-full arp=enabled \

```

### DIFF
--- a/plugins/modules/facts.py
+++ b/plugins/modules/facts.py
@@ -285,6 +285,11 @@ class Config(FactsBase):
     def populate(self):
         super(Config, self).populate()
         data = self.responses[0]
+
+        # remove datetime
+        data = re.sub(
+            r'^# [a-z0-9/][a-z0-9/]* [0-9:]* by RouterOS', r'# RouterOS', data
+        )
         if data:
             self.facts['config'] = data
 


heuels:
@qaxi, thanks for opening this PR! Could you also add a unit test for this change and a changelog fragment?

felixfontein (changes requested):
This is a breaking change. It need to be explicitly enabled by a module option.

(Docs on changelog fragments can be found here: https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to)

qaxi:
> This is a breaking change. It need to be explicitly enabled by a module option.

understand, will work on it, it needs more than such simple change

qaxi:
Just question for @felixfontein or @heuels 

what is better solution:

add new return value _ansible_net_config_idempotent_

```
    # config
    ansible_net_config:
      description: The current active config from the device
      returned: when config is configured
      type: str

    ansible_net_config_idempotent:
      description: The current active config from the device in minimal idempotent form
      returned: when config is configured
      type: str
```

or add option _idempotent_config_ with default _false_

```
option:
  idempotent_config:
    description:
      - When set C(true), config data will contain nonverbose config and will
        be idempotent between subsequent runs of facts module. 
    required: false
    default: false
```

felixfontein:
The problem with putting 'idempotent' in there is that it really needs to be idempotent. So if RouterOS ever starts adding something non-deterministic somewhere, we have to rush and filter it out / fix it somehow. That's potentially a very complex and complicated undertaking, and I'm not sure if we should commit on really doing it that way. Also you already mentioned another idempotency problem in #20, which would need to be covered as well for this to be idempotent, like the random MAC address you've mentioned, which would actually require us to parse and understand the config. I think that's not something that this module should do.

I see two options:
1. Have another module for retrieving only the config, and which tries to make it idempotent - would be more complex, or do a lot more calls, to try to eliminate all randomness.
2. Move that to a jinja2 filter plugin, which processes a config as a string. (That couldn't do multiple requests, so it would have to understand the config to a certain degree.)

qaxi:
There are two things

The date
- it is in _/export_ and _/export verbose_ output too 
- it is easy to filter it out
IMHO - it could be filtered from both outputs, but it is OK let current config mode untouched

The MAC randomness
- it is there because _/export verbose_ does show state of _unconfigured_ services with auto-generated parameters (https://wiki.mikrotik.com/wiki/Manual:Interface/OVPN mac-address ...)
- if you configure something _/export_ will show you just and only what you wrote ... so it is 'idempotent'

I understand your worries, so lets remove _idempotent_ and replace it with _verbose_ .

In this case I would prefer the second form - option

```
    # config
    ansible_net_config:
      description: The current active config from the device
      returned: when config is configured
      type: str

    ansible_net_config_nonverbose:
      description: The current active config from the device in minimal form
      returned: when config is configured
      type: str
```

or add option _verbose_config_ with default _true_

```
option:
  verbose_config:
    description:
      - When set C(false), config data will contain nonverbose config 
    required: false
    default: true
```

felixfontein:
Ok, if `/export` is fine, having `..._config_nonverbose` sounds good to me. It definitely should come with `version_added: 1.2.0` though (and a changelog fragment :) ). Do we want to remove the date (resp. first line) always, or make that configurable? Both would be fine for me. @heuels any preference?

heuels:
@felixfontein, I'd remove it unconditionally. If you want current date, there are more effective ways of getting it than `routeros_config`, right? 🙂

felixfontein:
@heuels good point! If someone wants the exact output of `/export`, they can just use the `community.routeros.command` module :)

PR #29, opened 2021-05-09
Clarify Windows (non-)support

felixfontein:
##### SUMMARY
Since this is not clear to every user, clarify in the README that this collection does not support Windows targets.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,8 +5,6 @@ Provides modules for [Ansible](https://www.ansible.com/community) to manage [Mik
 
 You can find [documentation for the modules and plugins in this collection here](https://ansible.fontein.de/collections/community/routeros/).
 
-Please note that this collection does **not** support Windows targets for the modules not using the `network_cli` connection.
-
 ## Tested with Ansible
 
 Tested with the current Ansible 2.9, ansible-base 2.10 and ansible-core 2.11 releases and the current development version of ansible-core. Ansible versions before 2.9.10 are not supported.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,6 +5,8 @@ Provides modules for [Ansible](https://www.ansible.com/community) to manage [Mik
 
 You can find [documentation for the modules and plugins in this collection here](https://ansible.fontein.de/collections/community/routeros/).
 
+Please note that this collection does **not** support Windows targets for the modules not using the `network_cli` connection.
```
heuels, line 8: Sorry, I can't seem to grasp the meaning. **Windows target** here is a host machine on which Ansible is being run, right?

felixfontein: Target is the machine where the module is run on, when using the regular connection plugin. When using the network_cli connection, the modules are run locally (and connect to a network device), so for them there's technically no need to document this. But the `api` module is a regular module and - when not `delegate_to: localhost` is added or `localhost` is used - will run on whatever target the user is using (which cannot be a RouterOS device, but for example a jump host in a network which has access to the RouterOS device). Basically the above says "we do not support Windows jump hosts for the `community.routeros.api` module".

I guess we need to formulate this more clearly though :)

heuels: Got it, thanks! Should we maybe just straight-up say it then, I think this is perfect:

> ### Known issues / edge cases
> Please note that `community.routeros.api` module does **not** support Windows jump hosts!

heuels: It's great! 👍🏻

felixfontein:
Zuul is currently broken, I'm therefore ignoring that negative check.

felixfontein:
@heuels thanks for reviewing!

PR #34, opened 2021-06-10
Disable use of deprecated Ciphers

jadyndev:
##### SUMMARY
Old ciphers shouldn't be used. Devices that use newer ciphers couldn't be reached as the cipher was locked to sslv3.

##### ISSUE TYPE

- Bugfix Pull Request

### DIFF
--- a/plugins/modules/api.py
+++ b/plugins/modules/api.py
@@ -451,7 +451,6 @@ class ROS_api_module:
                     conn_status["connection"]["port"] = port
                 ctx = ssl.create_default_context()
                 ctx.check_hostname = False
-                ctx.set_ciphers('ADH:@SECLEVEL=0')
                 api = connect(username=username,
                               password=password,
                               host=host,
--- a/changelogs/fragments/34-api-ciphers.yml
+++ b/changelogs/fragments/34-api-ciphers.yml
@@ -1,0 +1,2 @@
+bugfixes:
+- "api - when using TLS/SSL, remove explicit cipher configuration to insecure values, which also makes it impossible to connect to newer RouterOS versions (https://github.com/ansible-collections/community.routeros/pull/34)."


plugins/modules/api.py
```diff
@@ -451,7 +451,6 @@ def ros_api_connect(self, username, password, host, port, use_ssl):
                     conn_status["connection"]["port"] = port
                 ctx = ssl.create_default_context()
                 ctx.check_hostname = False
-                ctx.set_ciphers('ADH:@SECLEVEL=0')
```
felixfontein, line 454: @NikolayDachev was there a reason why you added this? Does it not work for you without this line? If yes, we should add a setting for it.

(We should also do that for the line above, `check_hostname`.)

felixfontein (commented):
Thanks for your contribution! Indeed, this isn't a good idea. I'm mainly wondering whether we should simply remove this line (as you suggested), or make this behavior configurable.

plugins/modules/api.py
felixfontein: I did some testing, with this line I could neither connect to my device when "TLS version" was set to "any", nor when it was set to "only v1.2". So if there's no explanation why this line is needed on some devices, let's just remove it. We can make it configurable if someone reports that it is really needed in some cases.

felixfontein:
@jadyndev can you please add a changelog fragment so we can merge this? For that, add a file `changelogs/fragments/34-api-ciphers.yml` with content:
```.yaml
bugfixes:
- "api - when using TLS/SSL, remove explicit cipher configuration to insecure values, which also makes it impossible to connect to newer RouterOS versions (https://github.com/ansible-collections/community.routeros/pull/34)."
```
(Or something similar; see the [guidelines](https://docs.ansible.com/ansible/devel/community/development_process.html#changelogs-how-to).)

felixfontein (approved):

heuels (approved):
@jadyndev, thanks for submitting this PR!
@felixfontein, thanks for reviewing!

felixfontein:
@jadyndev thanks for your work on this!
@heuels thanks for reviewing!

PR #37, opened 2021-06-20
Add options to control certificate validation for api module

felixfontein:
##### SUMMARY
Follow-up to https://github.com/ansible-collections/community.routeros/pull/34#discussion_r649076021.

This adds three options:
- `validate_certs` (default value `true`);
- `validate_cert_hostname` (default value `false`, for backwards compatibility);
- `ca_path` (same name as similar option in ansible/ansible#71979).

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
api

Issue #33, opened 2021-05-21
New task module for configuring


jplitza:
##### SUMMARY

Add a new task module that makes it easier to idempotently modify RouterOS config.

##### ISSUE TYPE
- Feature Idea

##### ADDITIONAL INFORMATION

Currently, it's a huge PITA to idemtpotently modify RouterOS config. My roles have RouterOS script snippets like this all over the place, to be executed via CLI:
```
:if ([/ip firewall filter find chain=forward action=fasttrack-connection] = "") do={/ip firewall filter add chain=forward action=fasttrack-connection connection-state=established,related ipsec-policy=out,none} else={/ip firewall filter set [find chain=forward action=fasttrack-connection] ipsec-policy=out,none}
```
This looks for a firewall rule in the forward chain with action "fasttrack-connection". If it finds such a rule, it sets "ipsec-policy=out,none" on it. If it doesn't find such a rule, it adds one.

Obviously it's quite impossible to catch all edge cases (especially ordering etc.), because RouterOS config is damn complex. But I'd imagine something like this:

```yaml
- community.routeros.config:
    path: /ip firewall filter
    attributes:
      chain: forward
      action: fasttrack-connection
      connection-state: established,related
      ipsec-policy: out,none
    match:
      - chain
      - action
```
One could also think about adding more parameters, like "place-before" (ideally accepting something like `[find ...]`), which would obviously only be used for adding, not for updating.

Or is there a more elegant way to do this?

felixfontein:
> Currently, it's a huge PITA to idemtpotently modify RouterOS config.

I definitely agree on that. This collection currentl only provides basic modules for talking with RouterOS devices, and nothing more fancy.

I'm not familiar with other networking collections, but I know that people did a lot of work for solving such things. So it's probably worth to investigate what other networking collections do before starting to implement something new :) AFAIK resource modules are currently a big thing. (Unfortunately I forgot the little I knew about what they are...)

jplitza:
> I'm not familiar with other networking collections, but I know that people did a lot of work for solving such things. So it's probably worth to investigate what other networking collections do before starting to implement something new :) AFAIK resource modules are currently a big thing. (Unfortunately I forgot the little I knew about what they are...)

Okay, so I'll just dump more of my thoughts about this approach:

[...]

Issue #32, opened 2021-05-21
Fail task on !trap from API


jplitza:
##### SUMMARY

It is currently impossible to catch a failed API command, because the TrapError from librouteros is specifically excluded from failing the module. (And also because the `errors` method incorrectly compares to `"False"` (string) instead of `False` (boolean))

IMO this is the sole advantage of the API over the CLI: That it communicates what is an error in a meaningful way.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

community.routeros.api

##### ADDITIONAL INFORMATION

```yaml

```

heuels:
cc @NikolayDachev

NikolayDachev:
> cc @NikolayDachev 

Thanks, will check this days

felixfontein:
> IMO this is the sole advantage of the API over the CLI: That it communicates what is an error in a meaningful way.

Another big advantage is that you don't have to handle funky CLI problems like #31 :)

NikolayDachev:
hey sorry I don't have physicals time to check this however can you please add example .. will help a lot

felixfontein:
I've started a fix in #39 btw.

felixfontein:
I think this is fixed by #39.

PR #39, opened 2021-06-26
Allow api module to fail

felixfontein:
##### SUMMARY
Right now, the module will not fail when `return_result` is called with `status=False`, since it looks for `status == 'False'` (string instead of boolean).

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
api

### DIFF
--- a/plugins/modules/api.py
+++ b/plugins/modules/api.py
@@ -458,7 +458,7 @@ class ROS_api_module:
             self.errors(e)
 
     def return_result(self, ch_status=False, status=True):
-        if status == "False":
+        if not status:
             self.module.fail_json(msg=to_native(self.result['message']))
         else:
             self.module.exit_json(changed=ch_status,
--- a/plugins/modules/api.py
+++ b/plugins/modules/api.py
@@ -266,6 +266,7 @@ import traceback
 LIB_IMP_ERR = None
 try:
     from librouteros import connect
+    from librouteros.exceptions import LibRouterosError
     from librouteros.query import Key
     HAS_LIB = True
 except Exception as e:
@@ -373,7 +374,7 @@ class ROS_api_module:
             for i in self.api_path:
                 self.result['message'].append(i)
             self.return_result(False, True)
-        except Exception as e:
+        except LibRouterosError as e:
             self.errors(e)
 
     def api_add(self):
@@ -382,7 +383,7 @@ class ROS_api_module:
             self.result['message'].append("added: .id= %s"
                                           % self.api_path.add(**param))
             self.return_result(True)
-        except Exception as e:
+        except LibRouterosError as e:
             self.errors(e)
 
     def api_remove(self):
@@ -390,7 +391,7 @@ class ROS_api_module:
             self.api_path.remove(self.remove)
             self.result['message'].append("removed: .id= %s" % self.remove)
             self.return_result(True)
-        except Exception as e:
+        except LibRouterosError as e:
             self.errors(e)
 
     def api_update(self):
@@ -401,7 +402,7 @@ class ROS_api_module:
             self.api_path.update(**param)
             self.result['message'].append("updated: %s" % param)
             self.return_result(True)
-        except Exception as e:
+        except LibRouterosError as e:
             self.errors(e)
 
     def api_query(self):
@@ -440,7 +441,7 @@ class ROS_api_module:
                     msg = msg + ' WHERE %s' % ' '.join(self.where)
                 self.result['message'].append(msg)
             self.return_result(False)
-        except Exception as e:
+        except LibRouterosError as e:
             self.errors(e)
 
     def api_arbitrary(self):
@@ -454,7 +455,7 @@ class ROS_api_module:
             for i in arbitrary_result:
                 self.result['message'].append(i)
             self.return_result(False)
-        except Exception as e:
+        except LibRouterosError as e:
             self.errors(e)
 
     def return_result(self, ch_status=False, status=True):
--- a/plugins/modules/api.py
+++ b/plugins/modules/api.py
@@ -468,7 +468,7 @@ class ROS_api_module:
     def errors(self, e):
         if e.__class__.__name__ == 'TrapError':
             self.result['message'].append("%s" % e)
-            self.return_result(False, True)
+            self.return_result(False, False)
         self.result['message'].append("%s" % e)
         self.return_result(False, False)
 
--- a/plugins/modules/api.py
+++ b/plugins/modules/api.py
@@ -460,7 +460,7 @@ class ROS_api_module:
 
     def return_result(self, ch_status=False, status=True):
         if not status:
-            self.module.fail_json(msg=to_native(self.result['message']))
+            self.module.fail_json(msg=self.result['message'])
         else:
             self.module.exit_json(changed=ch_status,
                                   msg=self.result['message'])
--- a/changelogs/fragments/39-api-fail.yml
+++ b/changelogs/fragments/39-api-fail.yml
@@ -1,6 +1,6 @@
-bugfixes:
+minor_changes:
 - >-
-  api - due to a programming error, the module never failed on errors.
+  api - due to a programming error, the module never failed on errors. This has now been fixed.
   If you are relying on the module not failing in case of idempotent commands
   (resulting in errors like ``failure: already have such address``), you need to adjust your roles/playbooks.
   We suggest to use ``failed_when`` to accept failure in specific circumstances,
--- a/changelogs/fragments/39-api-fail.yml
+++ b/changelogs/fragments/39-api-fail.yml
@@ -1,4 +1,4 @@
-minor_changes:
+breaking_changes:
 - >-
   api - due to a programming error, the module never failed on errors. This has now been fixed.
   If you are relying on the module not failing in case of idempotent commands
--- a/changelogs/fragments/39-api-fail.yml
+++ b/changelogs/fragments/39-api-fail.yml
@@ -1,2 +1,8 @@
 bugfixes:
-- "api - due to a programming error, the module never failed on errors (https://github.com/ansible-collections/community.routeros/pull/39)."
+- >-
+  api - due to a programming error, the module never failed on errors.
+  If you are relying on the module not failing in case of idempotent commands
+  (resulting in errors like ``failure: already have such address``), you need to adjust your roles/playbooks.
+  We suggest to use ``failed_when`` to accept failure in specific circumstances,
+  for example ``failed_when: "'failure: already have ' in result.msg[0]"``
+  (https://github.com/ansible-collections/community.routeros/pull/39).
--- a/changelogs/fragments/39-api-fail.yml
+++ b/changelogs/fragments/39-api-fail.yml
@@ -1,0 +1,2 @@
+bugfixes:
+- "api - due to a programming error, the module never failed on errors (https://github.com/ansible-collections/community.routeros/pull/39)."
--- a/tests/unit/plugins/modules/test_api.py
+++ b/tests/unit/plugins/modules/test_api.py
@@ -21,35 +21,23 @@ import json
 import pytest
 
 from ansible_collections.community.routeros.tests.unit.compat.mock import patch, MagicMock
-from ansible_collections.community.routeros.tests.unit.plugins.modules.utils import set_module_args, basic, AnsibleExitJson, AnsibleFailJson, ModuleTestCase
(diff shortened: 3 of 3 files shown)


felixfontein:
This took a bit longer, since there were some problems with the tests. These should be fixed now as well :)

NikolayDachev:
will check today

felixfontein:
(I rebased after merging #37 so the PR's branch contains that PR as well. I didn't make any changes to this PR's code.)

felixfontein:
Ok, this should either be merged in a bit more than a hour, or it has to skip this community.routeros release. Otherwise we might miss Ansible 4.2.0.

plugins/modules/api.py
```diff
@@ -454,11 +455,11 @@ def api_arbitrary(self):
             for i in arbitrary_result:
                 self.result['message'].append(i)
             self.return_result(False)
-        except Exception as e:
+        except LibRouterosError as e:
             self.errors(e)
 
     def return_result(self, ch_status=False, status=True):
```
NikolayDachev, line 461: From def errors we return True when we get TrapError, I do that with purpose. No easy way to do pre check via api for different tasks in ROS ! For this reason I ask for example in #32. 

for this approach (original code)
**Pros:** No need to handle **false positive** failers (we have such in ros)
**Cons:** Yes we will get fail as is excepted ! however for the  **false positive**  cases users must handle them via roles/playbooks 

For example: 
`    "msg": "['failure: already have such address']"
`
This happen when you try to add same ip address to interface .. first run will be successful, second run will fail !
Now I know this is my strange opinion, however this save overhand for such false positive cases like add ip address.

If we allow to FAIL in such case, the role / playbook must have SPECIFIC task to check if such address already exist end  etc .. However this is only for add ip what about all other basilion different tasks in ros. For each different task playook/role must have specific pre-check 

I hope you get my point way I do that in first place .. Over all I'm ok to make it fail (change must be done in  def errors() ) Probably additional module var to Fail or Not to Fail will be a good idea :)

    def errors(self, e):
        if e.__class__.__name__ == 'TrapError':
            self.result['message'].append("%s" % e)
            self.return_result(False, **True**)
        self.result['message'].append("%s" % e)
        self.return_result(False, False)

**NOTE: I'm really sorry for this late review :(**

NikolayDachev: When I think more, probably my opinion is wrong and we should make it to fail.

NikolayDachev: one more think: 

Is the correct approach to make it "Fail" -- I fully agree however this will make idempotent of the roles nightmare, the reason is because ros don't make a difference between foe example "wrong ip address mask" and "ip address already exist " (as example).
For this reason this module don't support check_mode 

May be I miss something and more research is needed ... 
I really don't know ...

NikolayDachev:
```
[admin@chr7b] > ip address/add address=192.168.12.3 netmask=255.255.256.0 interface=ether2
invalid value for argument netmask
[admin@chr7b] > ip address/add address=192.168.12.3 netmask=255.255.255.0 interface=ether2  
[admin@chr7b] > ip address/add address=192.168.12.3 netmask=255.255.255.0 interface=ether2 
failure: already have such address
```


same is via api module if we make it to fail
**first run:** ansible-playbook playbooks/ros_ip_addr_add.yml --limit chr7b --ask-vault-pass -vvvv

```
ok: [chr7b] => {
    "msg": {
        "changed": true,
        "failed": false,
        "msg": [
            "added: .id= *E"
        ]
    }
}
```

**second run:** ansible-playbook playbooks/ros_ip_addr_add.yml --limit chr7b --ask-vault-pass -vvvv

```
    "changed": false,
    "invocation": {
        "module_args": {
            "add": "broadcast= comment= copy-from= disabled=no netmask=255.255.255.0 network=192.168.192.0 address=192.168.192.20 interface=ether2",
            "ca_path": null,
            "cmd": null,
            "hostname": "chr7b",
            "password": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER",
            "path": "ip address",
            "port": null,
            "query": null,
            "remove": null,
            "ssl": "false",
            "tls": false,
            "update": null,
            "username": "admin",
            "validate_cert_hostname": false,
            "validate_certs": true
        }
    },
    "msg": "['failure: already have such address']"
}

```

**wrong mask**  ansible-playbook playbooks/ros_ip_addr_add.yml --limit chr7b --ask-vault-pass -vvvv -e "ros_ip_addr_netmask=255.255.25.256"

```
    "changed": false,
    "invocation": {
        "module_args": {
            "add": "broadcast= comment= copy-from= disabled=no netmask=255.255.25.256 network=192.168.192.0 address=192.168.192.20 interface=ether2",
            "ca_path": null,
            "cmd": null,
            "hostname": "chr7b",
            "password": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER",
            "path": "ip address",
            "port": null,
            "query": null,
            "remove": null,
            "ssl": "false",
            "tls": false,
            "update": null,
            "username": "admin",
            "validate_cert_hostname": false,
            "validate_certs": true
        }
    },
    "msg": "['invalid value for argument netmask']"
}

```

felixfontein: > Is the correct approach to make it "Fail" -- I fully agree however this will make idempotent of the roles nightmare, the reason is because ros don't make a difference between for example "wrong ip address mask" and "ip address already exist " .
> For this reason this module don't support check_mode
> 
> May be I miss something and more research is needed ...
> I really don't know ...

Well, that's always the problem with low-level modules like `ansible.builtin.command` and `community.routeros.command` - it's a non-trivial amount of work to make them do something idempotent. If a command executed on a host (for `ansible.builtin.command`) or a router (for `community.routeros.command`) fails, the module should fail, and the user should be able to decide how to react on the failure. Simplify making the module ignore errors by default is definitely not right, since the module has no idea whether an error was intended, acceptable, or not.

Ansible provides `failed_when` for tasks (https://docs.ansible.com/ansible/latest/user_guide/playbooks_error_handling.html#defining-failure) to allow role/playbook authors to define when a command really failed. This can be used to distinguish between "acceptable errors" and "unacceptable errors". In your above example, `failed_when` can be something like `result is success or (result is failed and 'failure: already have such address' in result.msg)`. That way, you can make the module invocation idempotent, and still report errors on wrong input.

In the long run, it would be great if there would be more advanced modules for such tasks, which know about what they run and how to handle it - see #33.

NikolayDachev: I agree in future will check with librouteros if is possible / usefully to utilize https://wiki.mikrotik.com/wiki/Manual:API#category somehow

jplitza: I propose to explicitly mention the case of "already have ..." errors that will now be reported in the changelog, since that could be considered a regression.

NikolayDachev:
Will try to check the test tonight

plugins/modules/api.py
felixfontein: The changelog fragment definitely needs to be extended. I'm wondering whether we should make this even more prominent, or whether this should even be a change for community.routeros 2.0.0.

NikolayDachev (approved):

NikolayDachev:
we can merge

NikolayDachev:
> Ok, this should either be merged in a bit more than a hour, or it has to skip this community.routeros release. Otherwise we might miss Ansible 4.2.0.

Since this is more like bug fix I gu

NikolayDachev:
closed by mistake

felixfontein:
@heuels @NikolayDachev @jplitza I've updated the changelog fragment. Please take a look! Also if you think that this shouldn't just show up in a patch release, but potentially a minor or even a major release, please say so! This will very likely break people's roles and playbooks, when they rely on errors not resulting in task failures. We have to decide how breaking this is (all collections stick to [semantic versioning](https://semver.org/)).

Basically our options are:
- Put this in a patch release (1.2.1). We should only do this if we think this is an important bugfix that - even though it breaks people's workflows - should get out fast.
- Put this in a minor release (1.3.0). This is only useful if someone installs the collection manually (i.e. outside of the Ansible package) and uses a version restriction like `< 1.2.0` to prevent new features.
- Put this in a major release (2.0.0). This indicates a breaking change. A consequence is that 2.0.0 will not be included in any Ansible 4.x.0 release, but only in Ansible 5.0.0 (to be released in mid November; see https://docs.ansible.com/ansible/devel/roadmap/COLLECTIONS_5.html).

What do you think?

NikolayDachev:
> @heuels @NikolayDachev @jplitza I've updated the changelog fragment. Please take a look! Also if you think that this shouldn't just show up in a patch release, but potentially a minor or even a major release, please say so! This will very likely break people's roles and playbooks, when they rely on errors not resulting in task failures. We have to decide how breaking this is (all collections stick to [semantic versioning](https://semver.org/)).
> 
> Basically our options are:
> 
> * Put this in a patch release (1.2.1). We should only do this if we think this is an important bugfix that - even though it breaks people's workflows - should get out fast.
> * Put this in a minor release (1.3.0). This is only useful if someone installs the collection manually (i.e. outside of the Ansible package) and uses a version restriction like `< 1.2.0` to prevent new features.
> * Put this in a major release (2.0.0). This indicates a breaking change. A consequence is that 2.0.0 will not be included in any Ansible 4.x.0 release, but only in Ansible 5.0.0 (to be released in mid November; see https://docs.ansible.com/ansible/devel/roadmap/COLLECTIONS_5.html).
> 
> What do you think?

1.3.0 look ok will give a some time for playbook/roles fixes

felixfontein:
1.3.0 will not mean that the release will happen later, it only changes the semantic of the release in terms of semantic versioning.

jplitza:
This definitely isn't a patch level change. I'm even tempted to say this is worth a new major version, since (roughly) a third of the collection changes a significant portion of its API (namely the return value, if you will). November isn't so far away, and this change isn't really pressing (it's been this way like forever, hasn't it?)

heuels:
For me it's also tempting to say it's a major-level release, but still I would stick with a minor-lever here.

NikolayDachev:
> 1.3.0 will not mean that the release will happen later, it only changes the semantic of the release in terms of semantic versioning.

Yes I know, what I try to point here was -> " This is only useful if someone installs the collection manually (i.e. outside of the Ansible package) and uses a version restriction like < 1.2.0 to prevent new features."

Other question: Why we should care about "ansible base" release date (except something new is pushed from it which we should implement). If I remember correctly this was the idea of small collections .. to not wait for "ansible base" release .. ?

felixfontein:
> Other question: Why we should care about "ansible base" release date (except something new is pushed from it which we should implement). If I remember correctly this was the idea of small collections .. to not wait for "ansible base" release .. ?

The ansible-core release date does not matter at all. What potentially matters is the Ansible release date. Ansible is the community distribution "with batteries included", which consists of ansible-core and a set of collections, including community.routeros. Creating a 2.0.0 release now means that the community.routeros version included in Ansible 4.x.0 will be frozen, except if we backport things and create further community.routeros 1.x.y releases. While this is possible (and not that much work), backporting features is more complicated, so the community.routeros version included in Ansible 4.x.0 will probably only get bugfixes, but no more new features.

Considering that we aren't publishing much features anyway, that's probably ok though.

NikolayDachev:
Well 1.3.0 ..

felixfontein:
I've relabelled the patch as minor_changes in 11280093b68a6593c772c0754c826ee6db394c0c. I'll try to discuss this with some people in the Ansible community next week to get some more feedback; right now I'm tending a bit to make this a major release.

NikolayDachev:
> I've relabelled the patch as minor_changes in 11280093b68a6593c772c0754c826ee6db394c0c. I'll try to discuss this with some people in the Ansible community next week to get some more feedback; right now I'm tending a bit to make this a major release.

Any news?

felixfontein:
From the feedback I got, this should really be a major release, since it is a breaking change in functionality. If that's fine with everyone, I'll adjust the changelog fragment accordingly.

felixfontein:
@heuels @NikolayDachev is this ok for you? If yes, I'll create a `stable-1` branch, merge this, and bump the next expected version in `main` to 2.0.0. (Then we can also think whether there are other breaking changes we want to do for a 2.0.0 release. Not sure if there is anything, but if there is something, that would be the time :) )

NikolayDachev:
I'm OK with it, thanks a lot for the effort

heuels:
Fine with me as well! Thanks for doing this.

felixfontein:
@NikolayDachev @jplitza @heuels thanks for your reviews and feedback!

Issue #44, opened 2021-08-23
[API] The comment breaks the work of the module


akimrx:
Of course, it was very funny to look for the cause of the error for so much time, but after looking at the source code, everything became clear.

Guys, we need to correct the division of the string into parameters, because the comment does not always consist of one word.

[The function that returns the error](https://github.com/ansible-collections/community.routeros/blob/a88f4c804b147136f9164e571408e90a649532e1/plugins/modules/api.py#L356)

[A function that incorrectly separates arguments](https://github.com/ansible-collections/community.routeros/blob/a88f4c804b147136f9164e571408e90a649532e1/plugins/modules/api.py#L380) and breaks down into `comment="some multiple string"`

**How to reproduce it?**  
Try to add a simple rule of the form:
```yaml
  community.routeros.api:
    hostname: "{{ mikrotik_host }}"
    username: "{{ mikrotik_user }}"
    password: "{{ mikrotik_password }}"
    path: ip firewall filter
    add: action=accept chain=forward comment='accept in ipsec policy' ipsec-policy=in,ipsec
```

The actual result will be an error when adding the rule, but by the way, this module does not understand its own errors at all and always returns OK, which is also funny and requires correction.

With the debug, I was able to see the true result of the execution and the essence of the "error":
```
    "msg": [
        "missing '=' after 'in'"
```

A few details:
```
ok: [localhost] => (item={'comment': 'accept in ipsec policy', 'id': 6, 'action': 'accept', 'chain': 'forward', 'options': 'ipsec-policy=in,ipsec'}) => {
    "ansible_loop_var": "item",
    "changed": false,
    "invocation": {
        "module_args": {
            "add": "action=accept chain=forward comment='accept in ipsec policy' ipsec-policy=in,ipsec",  <<< here
            "cmd": null,
            "hostname": "10.0.0.1",
            "password": "VALUE_SPECIFIED_IN_NO_LOG_PARAMETER",
            "path": "ip firewall filter",
            "port": null,
            "query": null,
            "remove": null,
            "ssl": false,
            "update": null,
            "username": "ansible"
        }
    },
    "item": {
        "action": "accept",
        "chain": "forward",
        "comment": "accept in ipsec policy",
        "ipsec-policy: "in,ipsec"
    },
    "msg": [
        "missing '=' after 'in'"  <<< here
```

felixfontein:
[...]

PR #45, opened 2021-08-23
fix(api): fixed a bug when splitting the argument string

akimrx:
##### SUMMARY
Closes https://github.com/ansible-collections/community.routeros/issues/44

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
`routeros_api`

### DIFF
--- a/plugins/modules/api.py
+++ b/plugins/modules/api.py
@@ -456,7 +456,7 @@ class ROS_api_module:
 
     def split_params(self, params):
         if not isinstance(params, str):
-            self.errors('Parameters can only be a string, received %s' % type(params))
+            raise AssertionError('Parameters can only be a string, received %s' % type(params))
         try:
             return split_routeros(params)
         except ParseError as e:
--- a/plugins/modules/api.py
+++ b/plugins/modules/api.py
@@ -367,9 +367,9 @@ class ROS_api_module:
         return dict
 
     def split_params(self, params):
-      if not isinstance(params, str):
-          self.errors('Parameters can only be a string, received %s' % type(params))
-      return shlex.split(params)
+        if not isinstance(params, str):
+            self.errors('Parameters can only be a string, received %s' % type(params))
+        return shlex.split(params)
 
     def api_add_path(self, api, path):
         api_path = api.path()
--- a/plugins/modules/api.py
+++ b/plugins/modules/api.py
@@ -313,7 +313,7 @@ class ROS_api_module:
                                         self.module.params['ca_path'],
                                         )
 
-        self.path = self.list_remove_empty(shlex.split(self.module.params['path']))
+        self.path = self.list_remove_empty(self.split_params(self.module.params['path']))
         self.add = self.module.params['add']
         self.remove = self.module.params['remove']
         self.update = self.module.params['update']
@@ -322,7 +322,7 @@ class ROS_api_module:
         self.where = None
         self.query = self.module.params['query']
         if self.query:
-            self.query = self.list_remove_empty(shlex.split(self.query))
+            self.query = self.list_remove_empty(self.split_params(self.query))
             try:
                 idx = self.query.index('WHERE')
                 self.where = self.query[idx + 1:]
@@ -366,6 +366,11 @@ class ROS_api_module:
                 dict[p[0]] = p[1]
         return dict
 
+    def split_params(self, params):
+      if not isinstance(params, str):
+          self.errors('Parameters can only be a string, received %s' % type(params))
+      return shlex.split(params)
+
     def api_add_path(self, api, path):
         api_path = api.path()
         for p in path:
@@ -381,7 +386,7 @@ class ROS_api_module:
             self.errors(e)
 
     def api_add(self):
-        param = self.list_to_dic(shlex.split(self.add))
+        param = self.list_to_dic(self.split_params(self.add))
         try:
             self.result['message'].append("added: .id= %s"
                                           % self.api_path.add(**param))
@@ -398,7 +403,7 @@ class ROS_api_module:
             self.errors(e)
 
     def api_update(self):
-        param = self.list_to_dic(shlex.split(self.update))
+        param = self.list_to_dic(self.split_params(self.update))
         if '.id' not in param.keys():
             self.errors("missing '.id' for %s" % param)
         try:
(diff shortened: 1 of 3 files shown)


akimrx:
[All tests passed in my repo](https://github.com/akimrx/community.routeros/actions), but you can check again

heuels (approved):
Looks good to me! 

Although I haven't touched Python in a while now, @felixfontein may be a better judge here 🙂

NikolayDachev:
@akimrx Thanks a lot for the fix 

Initial testing look ok, however I want to do more deep testing with arbitary cmd, query etc..

felixfontein:
Before discussing on whether this should we merged, we should really discuss the underlying questions in #44 which need to be answered before we can decide whether this PR actually provides a correct fix.

NikolayDachev:
@akimrx can you please merge after #47 is committed.

felixfontein:
@akimrx I've rebased your branch.

plugins/modules/api.py
```diff
@@ -312,7 +313,7 @@ def __init__(self):
                                         self.module.params['ca_path'],
                                         )
 
-        self.path = self.list_remove_empty(self.module.params['path'].split(' '))
+        self.path = self.list_remove_empty(self.split_params(self.module.params['path']))
```
felixfontein, line 316: I'm not sure, but shouldn't the path split be split by spaces? I.e.
```suggestion
        self.path = self.module.params['path'].split()
```
After all `'  a b   c'.split() == ['a', 'b', 'c']`, and the path cannot contain any escaping or strings AFAIK?

plugins/modules/api.py
```diff
@@ -365,6 +366,11 @@ def list_to_dic(self, ldict):
                 dict[p[0]] = p[1]
         return dict
 
+    def split_params(self, params):
+        if not isinstance(params, str):
+            self.errors('Parameters can only be a string, received %s' % type(params))
```
felixfontein, line 371:
```suggestion
```
Either this should result in an `AssertionError` raised (since it is an internal error, and not a user-facing error), or this shouldn't be checked at all.

felixfontein: Fixed in ec0c4a995bb8bb0746d01364bd5ee6892e8a2ece.

plugins/modules/api.py
felixfontein: @NikolayDachev what do you think?

felixfontein:
I've added https://github.com/ansible-collections/community.routeros/issues/44#issuecomment-917619520 + a changelog fragment to this PR in 3bc446c67acd2c57a026ff8bd18cf45806f25fc0.

NikolayDachev:
Thanks again for the work here,

note: We should update module documentation about strings with '""' for users which define it as variables  , no sure how to define it but 

ros_ip_addr_comment: '"my long string"' -> **correct** 
ros_ip_addr_comment: "'my long string'" -> wrong 
ros_ip_addr_comment: "my long string" -> wrong 
ros_ip_addr_comment: 'my long string' -> wrong

felixfontein:
It's probably a good idea to simplify this in some ways:
1. Provide Jinja2 filters to quote and unquote parameters.
2. Allow to pass in parameters as a dictionary to the `api` module, so you don't need to encode everything as a string first.

NikolayDachev:
Ideally we should be able to use | as this example without escape symbols 

{{ ros_backup_script_withes }} - work
{{ ros_backup_script_noes }} - we want to work as well (if is possible ! )

```- name: ros query example
  hosts: ros_testing
  gather_facts: no
  connection: local

  tasks:
  - name: add script
    vars:
      ros_backup_script_withes: '"
        :local host value=[/system identity get name];
        :local date value=[/system clock get date];
        :local day [ :pick $date 4 6 ];
        :local month [ :pick $date 0 3 ];
        :local year [ :pick $date 7 11 ];
        :local name value=($host.\"-\".$day.\"-\".$month.\"-\".$year);
        /system backup save name=$name;
        /export file=$name;
        /tool fetch address=192.168.1.1 user=ros password=\"PASSWORD\" mode=ftp dst-path=(\"/mikrotik/rsc/\".$name.\".rsc\") src-path=($name.\".rsc\") upload=yes;
        /tool fetch address=192.168.1.1 user=ros password=\"PASSWORD\" mode=ftp dst-path=(\"/mikrotik/backup/\".$name.\".backup\") src-path=($name.\".backup\") upload=yes;
        "'
      
      ros_backup_script_noes: |
        :local host value=[/system identity get name];
        :local date value=[/system clock get date];
        :local day [ :pick $date 4 6 ];
        :local month [ :pick $date 0 3 ];
        :local year [ :pick $date 7 11 ];
        :local name value=($host."-".$day."-".$month."-".$year);
        /system backup save name=$name;
        /export file=$name;
        /tool fetch address=192.168.1.1 user=ros password="PASSWORD" mode=ftp dst-path=("/mikrotik/rsc/".$name.".rsc") src-path=($name.".rsc") upload=yes;
        /tool fetch address=192.168.1.1 user=ros password="PASSWORD" mode=ftp dst-path=("/mikrotik/backup/".$name.".backup") src-path=($name.".backup") upload=yes;

    community.routeros.api:
      hostname: "{{ inventory_hostname }}"
      ssl: "false"
      password: "{{ vault_common_pwd }}"
      username: "admin"
      path: "system script"
      add:
        name="test"
        source={{ ros_backup_script_withes }}
    register: scriptout
    delegate_to: localhost
  
  - name: result ip address
    debug:
      msg: '{{ scriptout }}'

```

NikolayDachev:
> 2\. Allow to pass in parameters as a dictionary to the `api` module, so you don't need to encode everything as a string first.

well the idea is to be close as ros cmd, but both cases are good idea as well !

NikolayDachev (approved):

felixfontein:
I agree, let's merge this and then work on improvements in further PRs.

felixfontein:
@akimrx thanks for your contribution!
@heuels @NikolayDachev thanks for reviewing!

felixfontein:
@NikolayDachev with #53, you can do
```
  - community.routeros.api:
      hostname: "{{ inventory_hostname }}"
      ssl: "false"
      password: "{{ vault_common_pwd }}"
      username: "admin"
      path: "system script"
      add:
        name="test"
        source={{ ros_backup_script_noes | community.routeros.quote_argument_value }}
    register: scriptout
    delegate_to: localhost
```

PR #59, opened 2021-11-25
Also collect coverage during sanity tests

felixfontein:
##### SUMMARY
Enables coverage for sanity tests. This is somewhat experimental since I'm not 100% sure what the result will be. From glancing at the code, I *think* this mostly affects the import sanity test, so that some of the code which handles failing imports will show up in coverage reports.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
CI


felixfontein:
The decrease in coverage is probably due to a file which hasn't been touched by the tests at all and thus didn't show up in coverage at all. So the main result is that coverage is now more honest.

Issue #86, opened 2022-04-28
community.routeros.facts fails for some devices


merlin-vrn:
##### SUMMARY
`community.routeros.facts` module fails with cryptic Python error message on some managed targets. I don't know how these targets are different from other targets; a help is needed to help me to investigate this problem. The "reduced" facts collection (only `hardware`) doesn't seem to fail.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
`community.routeros.facts` module

##### ANSIBLE VERSION

```paste below
ansible [core 2.12.4]
  config file = /home/merlin/ansible/rterminal-ansible-default/trunk/routeros/ansible.cfg
  configured module search path = ['/home/merlin/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/merlin/ansible/lib/python3.9/site-packages/ansible
  ansible collection location = /home/merlin/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/merlin/ansible/bin/ansible
  python version = 3.9.9 (main, Jan 29 2022, 08:46:33) [GCC 11.2.0]
  jinja version = 3.1.1
  libyaml = True
```

##### COLLECTION VERSION

```paste below
Collection         Version
------------------ -------
community.routeros 2.0.0  
```

I also tried the latest version from Github, got exactly same error message (up to apparently random strings).

##### CONFIGURATION

```paste below
DEFAULT_HOST_LIST(/home/merlin/ansible/rterminal-ansible-default/trunk/routeros/ansible.cfg) = ['/home/merlin/ansible/rterminal-ansible-default/trunk/routeros/hosts']
```

##### OS / ENVIRONMENT

Controller node: `Linux muon 5.15.23-gentoo-x86_64 #1 SMP PREEMPT Tue Feb 15 09:57:12 MSK 2022 x86_64 Intel(R) Core(TM) i5-3470 CPU @ 3.20GHz GenuineIntel GNU/Linux`
Ansible is running in virtualenv, it was installed right as recommended in the official docs.

Managed target: 2011UiAS-2HnD, RouterOS version 6.47.4. It shows less than 20% of CPU load and 91.4MiB of free memory.

Another target where it appeared: 962UiGS-5HacT2HnT, 6.49.2, which has low CPU load and much free memory available too.

And another one, RB4011iGS+, 6.43.4.

##### STEPS TO REPRODUCE

```yaml
ansible gw -m community.routeros.facts
```

[...]

Issue #62, opened 2021-12-17
Extra lines in stdout


1rV1N-git:
##### SUMMARY
When I send a command to the router I get a response from the router with extra lines. 
![image](https://user-images.githubusercontent.com/34376228/146521286-fac11c9e-886a-46ad-88f8-9f32dd431435.png)

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
routeros

##### ANSIBLE VERSION

```paste below
ansible 2.9.6
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/wsl/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.8.10 (default, Sep 28 2021, 16:10:42) [GCC 9.3.0]
```

##### COLLECTION VERSION

```paste below

```

##### CONFIGURATION

```paste below
[defaults]
stdout_callback = yaml
host_key_checking = False
retry_files_enabled = False
inventory = hosts.yaml
[persistent_connection]
connect_timeout = 30
command_timeout = 30
```

##### STEPS TO REPRODUCE

```yaml
- name: check user is exist
  hosts: mikrotik
  gather_facts: false
  tasks:
    - name: check user is exist
      community.routeros.command:
        commands: 
          - if ([/user find name=admin] !="") do={:put "exist"}
      register: exist

    - debug:
        var: exist
```

##### EXPECTED RESULTS
I want to get only output from line 3

##### ACTUAL RESULTS
I got 3 lines. First-line eq first letter of command, middle-lines = empty line(empty lines can be more than one), last line eq correct output.  

```paste below

```

1rV1N-git:
if i send the command with a space at the beginning it is ok
```yaml
        commands: 
          - ' if ([/user find name=admin] !="") do={:put "exist"}'
```

felixfontein:
Maybe it's related to https://docs.ansible.com/ansible/devel/collections/community/routeros/docsite/ssh-guide.html#important-notes; have you tried adding something like `+cet512w` to the RouterOS admin username?

1rV1N-git:
I already tried it

felixfontein:
Maybe it's related to how prompts are detected in the SSH output. If that's the case, this will not be simple to fix.

1rV1N-git:
it is sad. Maybe you will try. Please

[...]

PR #88, opened 2022-05-01
Add api_facts module

felixfontein:
##### SUMMARY
Similar to the `facts` module, but uses the API instead of SSH. Inspried by the problems in #86 and #62. I also moved some common code from the `api` module to a new module_utils and a new docs fragment.

This does not support the `config` information yet, I wasn't able to (quickly) figure out how to actually get them with the API. If anyone knows how to do that, I'd happily add it :)

The code of the module is very similar to the `facts` module. ~~It can be simplified a lot, might do that later.~~

@heuels @NikolayDachev what do you think?

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
api_facts


NikolayDachev:
I will check next week, I don't have a pshisical time to do that early

plugins/modules/api_facts.py
```diff
@@ -0,0 +1,483 @@
...
+- name: Do not collect hardware facts
+  community.routeros.api_facts:
+    hostname: 192.168.88.1
+    username: admin
+    password: password
+    gather_subset:
+      - "!hardware"
+"""
+
+RETURN = """
```
NikolayDachev, line 65: RETURN is almost the same as facts.py, can we use doc_fragments ?!

```
RETURN =  """ diff api_facts.py facts.py 


59a60,75
>     # config
>     ansible_net_config:
>       description: The current active config from the device.
>       returned: I(gather_subset) contains C(config)
>       type: str
> 
>     ansible_net_config_nonverbose:
>       description:
>         - The current active config from the device in minimal form.
>         - This value is idempotent in the sense that if the facts module is run twice and the device's config
>           was not changed between the runs, the value is identical. This is achieved by running C(/export)
>           and stripping the timestamp from the comment in the first line.
>       returned: I(gather_subset) contains C(config)
>       type: str
>       version_added: 1.2.0
> 
103c119
< """
---
> """
\ No newline at end of file
```

NikolayDachev (commented):
api - those are ok  
- doc_fragments
- module_utils 
- module

**api_facts.py** - I never used facts ( the old one) before :) so I want to do some testing on api_facts,  please give me some time to play with it . 

testing -  **fake_api** \m/ ... 👍 

Also I'm not sure what you mean with "This does not support the config information yet" probably I can help ?

plugins/modules/api_facts.py
felixfontein: There is a POC for that https://github.com/ansible/ansible/pull/72635 but it will take some more time until it gets merged. And then it takes even more time until the at that point current popularly used ansible-core versions all support it. So unfortunately not yet :)

felixfontein:
> Also I'm not sure what you mean with "This does not support the config information yet" probably I can help ?

The `facts` module also returns the result of `/export` and `/export verbose`. I've tried into looking to get the `api_facts` to export that as well, but it turns out that while you *can* run the `/export` command (using https://librouteros.readthedocs.io/en/3.2.0/path.html#arbitrary-command), its output isn't returned. Which makes it pretty useless :) What you *can* do is run it and let it write its output to a file. But then you would have to retrieve the contents of that file - probably with `scp`? - and afterwards clean the file. (And pick a filename which doesn't already exist in the first place.) That's way too much complication for this module IMO.

If someone wants to get the output of `/export` they'd have to do it "manually" with the `api` module (and `ansible.netcommon.net_get` or something like that).

NikolayDachev:
> > Also I'm not sure what you mean with "This does not support the config information yet" probably I can help ?
> 
> The `facts` module also returns the result of `/export` and `/export verbose`. I've tried into looking to get the `api_facts` to export that as well, but it turns out that while you _can_ run the `/export` command (using https://librouteros.readthedocs.io/en/3.2.0/path.html#arbitrary-command), its output isn't returned. Which makes it pretty useless :) What you _can_ do is run it and let it write its output to a file. But then you would have to retrieve the contents of that file - probably with `scp`? - and afterwards clean the file. (And pick a filename which doesn't already exist in the first place.) That's way too much complication for this module IMO.
> 
> If someone wants to get the output of `/export` they'd have to do it "manually" with the `api` module (and `ansible.netcommon.net_get` or something like that).

Yes `/export` not working, will think on it, I know I try the same before and there was a particular reason but I don't remember it. We have a option to use REST api + ansible.uri  (not sure if have /export) but this will work only with ros7

NikolayDachev:
I do some testing, l really like the results :) 

At this stage I think can be approved except we want to try solve ```/export``` somehow ?

felixfontein:
Since `/export` seems more complicated (and not like something we can easily solve very soon), let's merge it now without that, and maybe try to add it later. In the original `facts` module `config` wasn't included by default anyway, so users had to explicitly enable it.

If that's ok I'll merge.

NikolayDachev (approved):
ready to be merge

NikolayDachev:
> Since `/export` seems more complicated (and not like something we can easily solve very soon), let's merge it now without that, and maybe try to add it later. In the original `facts` module `config` wasn't included by default anyway, so users had to explicitly enable it.
> 
> If that's ok I'll merge.

``

NikolayDachev:
sorry by mistake click close :(

felixfontein:
@NikolayDachev no problem, no damage done :) Thanks a lot for testing this!

PR #95, opened 2022-05-24
Add encoding option to API modules

felixfontein:
##### SUMMARY
Fixes one of the issues mentioned in https://github.com/ansible-collections/community.routeros/pull/91#issuecomment-1134860696. Note that `encoding='UTF-8'` doesn't properly work without https://github.com/luqasz/librouteros/pull/131.

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
API modules


NikolayDachev:
side note: probably we should add few words for the new modules in README

felixfontein:
> side note: probably we should add few words for the new modules also for `encoding:` in git README

Good point, I'll write something in a follow-up PR.

felixfontein:
@NikolayDachev thanks for reviewing this!

PR #99, opened 2022-06-29
Add docs PR workflow that a) builds docs and b) show the docs diff that a PR ceates

felixfontein:
##### SUMMARY
This workflow is also used in community.general and community.docker and works fine there.

Unfortunately this workflow won't run in the PR, *but* once this is merged I'll create a new PR so we can see whether it passes. There might be some things to fix (which I will do in that PR), and once that's fixed it should work fine.

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
CI


felixfontein:
CC @samccann @briantist

PR #111, opened 2022-09-12
[ON HOLD] Install using ansible-galaxy via git instead of cloning directly

felixfontein:
##### SUMMARY
Install using ansible-galaxy via git instead of cloning directly.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
CI


felixfontein:
Looks like this only works once we drop support for Ansible 2.9...

felixfontein:
Integrated into #318.

PR #128, opened 2022-11-13
API update

therfert:
##### SUMMARY
- Suport for API fields that can be disabled and have a default at the same time.
- Support for API path: `` interface gre``

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
- api_modify
- api_info

##### ADDITIONAL INFORMATION
- The code makes sure the default value is used in case the field isn't specified in the ``data`` and ``handle_entries_content`` isn't ``ignore``. It also covers a situation when the field is currently disabled in ROS.
- It will disable the field only when it's explicitly defined (as empty value or key with exclamation mark prefix). 
- See ``interface gre`` field ``keepalive`` as an example where this is required.


therfert:
Hi @felixfontein! Could you please re-run the checks? (it failed because of some network issues again). Thanks

felixfontein (commented):
I don't see how `default` and `can_disable` can be used together. If there is a default and the value is not returned by the API, the modules treat the value of that field as the default value. If the field can be disabled and is not returned by the API, it is treated as disabled. If a field both has a default and is specified as 'can_disable', how should the module now handle the case that the value is not returned by the API?

therfert:
> I don't see how `default` and `can_disable` can be used together. If there is a default and the value is not returned by the API, the modules treat the value of that field as the default value. If the field can be disabled and is not returned by the API, it is treated as disabled. If a field both has a default and is specified as 'can_disable', how should the module now handle the case that the value is not returned by the API?

See the ``keepalive`` field of the of an item in ``interface gre``.
When the item is created, the default value is used and that's what's returned by the API.
However the field can be explicitly disabled and then the API returns the following value: ``!keepalive``

felixfontein:
Ok, I understand now why it is necessary.

I want to to through the other code in more depth though to make sure we're not forgetting to change a place which would be affected by this as well.

therfert:
> Ok, I understand now why it is necessary.
> 
> I want to to through the other code in more depth though to make sure we're not forgetting to change a place which would be affected by this as well.

Thanks!

therfert:
I added ``interface eoip`` path support into this PR as it requires the same feature for the ``keepalive`` field.

felixfontein:
I think you need the following change as well if such paths should ever be used in the tests:
```diff
diff --git a/tests/unit/plugins/modules/fake_api.py b/tests/unit/plugins/modules/fake_api.py
index b17c088..f50c996 100644
--- a/tests/unit/plugins/modules/fake_api.py
+++ b/tests/unit/plugins/modules/fake_api.py
@@ -120,7 +120,7 @@ class Or(object):
 
 def _normalize_entry(entry, path_info):
     for key, data in path_info.fields.items():
-        if key not in entry and data.default is not None:
+        if key not in entry and data.default is not None and not data.can_disable:
             entry[key] = data.default
         if data.can_disable:
             if key in entry and entry[key] in (None, data.remove_value):
```
Besides that, it looks good to me!

therfert:
@felixfontein done. btw. I also thought about the tests, but I had never done that and didn't really know where to start. (I'm not a developer :D )

felixfontein (approved):

felixfontein:
Thanks again for yet another contribution :)

PR #143, opened 2022-12-26
Add wireguard interface and peers

cosandr:
##### SUMMARY

Adds support for configuring wireguard interfaces and peers.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

* api_modify
* api_info

##### ADDITIONAL INFORMATION

Missing tests and equally as important, `interface wireguard` editing doesn't behave as expected (misconfiguration error, see below). The API obfuscates the value of `private-key` so the diff is broken, for example:

```
--- before
+++ after
@@ -6,7 +6,7 @@
             "listen-port": 13231,
             "mtu": 1420,
             "name": "WG1",
-            "private-key": "*****"
+            "private-key": "mLh5K+/uJmZJ1zqarPYh0OyBaCgI1n+xAUBvsPBC+3I="
         }
     ]
 }
```

This happens even if that private key is already configured on the router. I was unable to find documentation on how to get the API to return sensitive information like this.

UPDATE: This is fixed by adding the "sensitive" policy to the group used to connect.

```
/user/group> set ansible policy=read,write,api,sensitive
```

After that, the private-key is returned and diff works as expected.

### DIFF
--- a/changelogs/fragments/143-add-wireguard.yml
+++ b/changelogs/fragments/143-add-wireguard.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - api_modify, api_info - support API paths ``interface wireguard``, ``interface wireguard peers`` (https://github.com/ansible-collections/community.routeros/pull/143).
--- a/plugins/module_utils/_api_data.py
+++ b/plugins/module_utils/_api_data.py
@@ -1150,6 +1150,33 @@ PATHS = {
             'verify-client-certificate': KeyInfo(default='no'),
         },
     ),
+    ('interface', 'wireguard'): APIData(
+        fully_understood=True,
+        primary_keys=('name', ),
+        fields={
+            'comment': KeyInfo(can_disable=True, remove_value=''),
+            'disabled': KeyInfo(default=False),
+            'listen-port': KeyInfo(),
+            'mtu': KeyInfo(default=1420),
+            'name': KeyInfo(),
+            'private-key': KeyInfo(),
+        },
+    ),
+    ('interface', 'wireguard', 'peers'): APIData(
+        fully_understood=True,
+        primary_keys=('public-key', 'interface'),
+        fields={
+            'allowed-address': KeyInfo(),
+            'comment': KeyInfo(can_disable=True, remove_value=''),
+            'disabled': KeyInfo(default=False),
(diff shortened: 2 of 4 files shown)


felixfontein (commented):
Thanks for your contribution! Can you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Thanks.

felixfontein:
> Missing tests and equally as important, `interface wireguard` editing doesn't behave as expected.

Hmm, that's unfortunate. We probably should document this behavior, and/or have some special code to handle this in a sane (and preferably user-configurable) way.

cosandr:
> Hmm, that's unfortunate. We probably should document this behavior, and/or have some special code to handle this in a sane (and preferably user-configurable) way.

I feel like there has to be a way to actually get this information from the API, I can try asking them directly. In the CLI the private key is hidden unless using export (without hide-sensitive at least).

If there isn't, we could add a `sensitive` and/or `force` attribute to KeyInfo, without force it would ignore changes.

cosandr:
@felixfontein I've figured out the issue with private-key and edited the description, it was my fault for not giving my API user enough permissions. Let me know if there's anything more I should do to get this merged, thanks!

changelogs/fragments/143-add-wireguard.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - api_modify, api_info - support API paths ``interface wireguard``, ``interface wireguard peers``
```
felixfontein, line 2:
```suggestion
  - api_modify, api_info - support API paths ``interface wireguard``, ``interface wireguard peers`` (https://github.com/ansible-collections/community.routeros/pull/143).
```

felixfontein (commented):
Great to hear! I have one small comment for the changelog fragment, besides that it looks good!

changelogs/fragments/143-add-wireguard.yml
cosandr: Thanks, fixed.

felixfontein (approved):

felixfontein:
@cosandr thanks for your contribution!

PR #160, opened 2023-03-14
Add "handle_stratified" support

phox142:
##### SUMMARY
The goal of this change is to permit to `handle_absent_entries=remove` and `ensure_order=True` to work on subset based on `stratify_keys`  
For instance if the parameter `handle_stratified` is set to `True` when working on firewall rules, only the rules in the same `chain` will be affected.  The other chains will not be adapted.
##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
api module

### DIFF
--- a/plugins/modules/api_modify.py
+++ b/plugins/modules/api_modify.py
@@ -551,10 +551,10 @@ def sync_list(module, api, path, path_info):
     handle_absent_entries = module.params['handle_absent_entries']
     handle_entries_content = module.params['handle_entries_content']
     handle_stratified = module.params['handle_stratified']
-    
-    if handle_stratified and path_info.stratify_keys is None: 
+
+    if handle_stratified and path_info.stratify_keys is None:
         module.fail_json('For this path, handle_stratified cannot be True')
-        
+
     if handle_absent_entries == 'remove':
         if handle_entries_content == 'ignore':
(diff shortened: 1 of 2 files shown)


plugins/modules/api_modify.py
```diff
@@ -228,6 +228,11 @@
       - remove
       - remove_as_much_as_possible
     default: ignore
+  handle_stratified:
+    description:
+        - TODO
+    type: bool
+    default: false
```
felixfontein, line 235: How about instead allowing to specify which keys (for stratification) to consider, so that only entries for these keys are accepted by the module and that only entries for these keys are considered?

irrwitzer42:
@phox142  Thanks so much for implementing  this! Looking forward to seeing this merged soon!

My use case are address lists - I have common ones on every router and I have router specific ones. At the moment, the specific ones would be removed on every playbook run. 

Thanks again!

felixfontein:
I created another PR with a more flexible approach (basically what I mentioning in https://github.com/ansible-collections/community.routeros/pull/160/files/0aaa34e0d71af860c83c379e9dcbd9bb338c863d#diff-b46a57970bad777c71fc3d869bef8246fd76e6ec947fb99012b34d1f0f78b130) in #305.

ewysong:
@felixfontein awesome! Looking forward to it being merged :smiley:

felixfontein:
@phox142 thanks for working on this!

PR #203, opened 2023-08-21
Finalize fields for "interface wireless security-profiles"

hansmi:
##### SUMMARY
A few fields were missing for `interface wireless security-profiles` (e.g. `disabled`). `default` on the other hand is a built-in property which can't be modified.

Mark the `interface wireless security-profiles` path as fully understood to enable its use in Ansible tasks.

##### ISSUE TYPE
- Bugfix/Feature Pull Request

##### COMPONENT NAME
community.routeros


felixfontein:
Thanks for updating this path! It's one of the ones I added initially (based on some stuff I managed to extract from the CLI), but since I don't use it I didn't inspect is closer. Back then I was still hoping there would be a nicer / halfway automatic way to get information on parameters for the paths from RouterOS somehow, which I haven't managed to find so far.

felixfontein (approved):

felixfontein:
Thanks for your contribution!

Issue #208, opened 2023-08-30
SNMP configuration not more working since RouterOS 7.10


phox142:
##### SUMMARY
Since version 7.10 Mikrotik added  snmp "engine-id-suffix" setting and set "engine-id" as read-only property;  
Therefore the api disallow  the parameter "engine-id" in the configuration calls and "api_update" does not work anymore as it always put a "empty" value in the call.
I tried different solutions to adapt the code but I don't find any backward compatible solution as the only solution I found is to remove this field from "_api_data.py"

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
- api_update
- api_data

##### ANSIBLE VERSION

```paste below
ansible [core 2.15.2]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/***/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  ansible collection location = /home/***/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0] (/usr/bin/python3)
  jinja version = 3.0.3
  libyaml = True
```

##### COLLECTION VERSION
```paste below
2.9.0
```

##### CONFIGURATION
```paste below
CONFIG_FILE() = /etc/ansible/ansible.cfg
DEFAULT_JINJA2_EXTENSIONS(/etc/ansible/ansible.cfg) = jinja2.ext.do
```

##### OS / ENVIRONMENT
Ubuntu 22.04

##### STEPS TO REPRODUCE

```yaml
- name: Setup SNMP
  community.routeros.api_modify:
    path: snmp
    handle_entries_content: remove
    data: 
      - enabled: true
        contact: 'contact@test.com'
        location': location
        trap-target': 'x.y.w.z'
        trap-community': 'comm'
        trap-version': 3
        trap-generators': 'interfaces,start-trap,temp-exception'
        trap-interfaces': 'ens01'
        engine-id-suffix: 123456789
```

##### EXPECTED RESULTS
Possibility to configure SNMP without error ;)

##### ACTUAL RESULTS

```paste below
"Error while modifying: unknown parameter engine-id"
```

felixfontein:
This was about to happen sooner or later... I'll see whether I can add some code to handle that.

felixfontein:
[...]

PR #209, opened 2023-08-31
Allow to differ on API paths based on RouterOS version (1/2)

felixfontein:
##### SUMMARY
Adresses the underlying problem in #208 and makes it possible to fix it.

- [x] Allow to provide different path definitions for API versions
- [ ] Allow to adjust parameters of a path for API versions (more granular) - will follow in another PR

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
API data


felixfontein:
I decided to do the second step in a follow-up PR. Since this PR does change the whole `_api_data.py` file's structure it's better to get it merged as soon as possible to not create too many conflicts with other contributions.

plugins/module_utils/_api_data.py
```diff
@@ -9,8 +9,67 @@
...
+                    break
+
+    def provide_version(self, version):
+        if not self.needs_version:
+            return self.unversioned.fully_understood
+        api_version = LooseVersion(version)
+        for other_version, comparator, unversioned in self.versioned:
+            if other_version == '*' and comparator == '*':
+                self.unversioned = unversioned
+                return self.unversioned.fully_supported
```
hansmi, line 58: I think the returned property should be `fully_understood`. There's no `fully_support`.

> AttributeError: 'VersionedAPIData' object has no attribute 'fully_supported'

PR #211, opened 2023-09-01
Add support for the "user" path

hansmi:
##### SUMMARY
Make it possible to manage users via the `user` path.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
community.routeros

PR #213, opened 2023-09-02
Implement basic support for read-only and write-only keys in api_info and api_modify

felixfontein:
##### SUMMARY
Ref: https://github.com/ansible-collections/community.routeros/pull/211/files#r1313462457

CC @hansmi - I didn't really have time to test this yet, but if you already want to try it out: enjoy ;-)

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
API data
api_info
api_modify

### DIFF
--- a/changelogs/fragments/213-read-write-only.yml
+++ b/changelogs/fragments/213-read-write-only.yml
@@ -1,0 +1,3 @@
+minor_changes:
+  - "api_info - add new ``include_read_only`` option to select behavior for read-only values. By default these are not returned (https://github.com/ansible-collections/community.routeros/pull/213)."
+  - "api_modify - add new ``handle_read_only`` and ``handle_write_only`` options to handle the module's behavior for read-only and write-only fields (https://github.com/ansible-collections/community.routeros/pull/213)."
--- a/plugins/modules/api_info.py
+++ b/plugins/modules/api_info.py
@@ -234,6 +234,13 @@ options:
     type: bool
     default: false
     version_added: 2.4.0
+  include_read_only:
+    description:
+      - Whether to include read-only fields.
+      - By default, they are not returned.
+    type: bool
+    default: false
+    version_added: 2.10.0
 seealso:
   - module: community.routeros.api
   - module: community.routeros.api_facts
@@ -313,6 +320,7 @@ def main():
         hide_defaults=dict(type='bool', default=True),
         include_dynamic=dict(type='bool', default=False),
         include_builtin=dict(type='bool', default=False),
+        include_read_only=dict(type='bool', default=False),
     )
     module_args.update(api_argument_spec())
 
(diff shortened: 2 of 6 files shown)


changelogs/fragments/213-read-write-only.yml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - "api_info - add new ``include_read_only`` option to select behavior for read-only values. By default these are not returned (https://github.com/ansible-collections/community.routeros/pull/213)."
```
hansmi, line 2: The documentation says `default: true` for `include_read_only`. I think the word `not` should be dropped here.

plugins/modules/api_info.py
```diff
@@ -234,6 +234,13 @@
     type: bool
     default: false
     version_added: 2.4.0
+  include_read_only:
+    description:
+      - Whether to include read-only fields.
+      - By default, they are not returned.
```
hansmi, line 240: This line contradicts the default value of `true`.

changelogs/fragments/213-read-write-only.yml
felixfontein: I switched everything to `false` except the documentation (that's also why the sanity test fails); I've updated that one instead.

Or do you think it's better to have the default as `true`? I was thinking about this a bit, and eventually decided to use the same default for all `include_*` options. But for this one, it could also make sense to have `true` as the default.

hansmi: I thought including read-only values by default is the more sensible choice for `api_info`, but you have a point with consistency with the other `include_*` options. Let's leave it as false.

felixfontein:
I did some testing with a modified version of #211:
```diff
diff --git a/plugins/module_utils/_api_data.py b/plugins/module_utils/_api_data.py
index 9dbe2a2..b92c611 100644
--- a/plugins/module_utils/_api_data.py
+++ b/plugins/module_utils/_api_data.py
@@ -3231,6 +3231,22 @@ PATHS = {
             },
         ),
     ),
+    ('user',): APIData(
+        unversioned=VersionedAPIData(
+            primary_keys=('name', ),
+            fully_understood=True,
+            fields={
+                'address': KeyInfo(),
+                'comment': KeyInfo(can_disable=True, remove_value=''),
+                'disabled': KeyInfo(default=False),
+                'expired': KeyInfo(read_only=True),
+                'group': KeyInfo(),
+                'last-logged-in': KeyInfo(read_only=True),
+                'name': KeyInfo(),
+                'password': KeyInfo(write_only=True),
+            },
+        ),
+    ),
     ('user', 'aaa'): APIData(
         unversioned=VersionedAPIData(
             single_value=True,
```
and it works fine (with the last change I pushed).

felixfontein:
@hansmi thanks for reviewing this!

PR #220, opened 2023-09-18
Support routing id and bgp connection

cosandr:
##### SUMMARY

This PR adds support for `routing id` and `routing bgp connection`.

Neither of these appear to have primary keys, the router accepts duplicates.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

* `api_info`
* `api_modify`

##### ADDITIONAL INFORMATION

Note that `routing bgp connection` has a lot more options than I've included in this PR.

<details>
<summary>For example on my ROS 7.10 device</summary>

  ```
  /routing/bgp/connection> add
  
  add-path-out --
  address-families --
  as --
  cisco-vpls-nlri-len-fmt --
  cluster-id --
  comment -- Short description of the item
  connect --
  copy-from -- Item number
  disabled -- Defines whether item is ignored or used
  hold-time --
  input.accept-communities --
  input.accept-ext-communities --
  input.accept-large-communities --
  input.accept-nlri --
  input.accept-unknown --
  input.affinity --
  input.allow-as --
  input.filter --
  input.ignore-as-path-len --
  input.limit-process-routes-ipv4 --
  input.limit-process-routes-ipv6 --
  keepalive-time --
  listen --
  local.address --
  local.port --
  local.role --
  local.ttl --
  multihop --
  name --
  nexthop-choice --
  output.affinity --
  output.as-override --
  output.default-originate --
  output.default-prepend --
  output.filter-chain --
  output.filter-select --
  output.keep-sent-attributes --
  output.network --
  output.no-client-to-client-reflection --
  output.no-early-cut --
  output.redistribute --
  output.remove-private-as --
  remote.address --
  remote.allowed-as --
  remote.as --
  remote.port --
  remote.ttl --
  router-id --
  routing-table --
  save-to --
  tcp-md5-key --
  templates --
  use-bfd --
  vrf --
  ```

</details>

Let me know if you think I should add those as well.

PS. I had to manually update the docs, the script is removing a bunch of stuff and not adding my new paths.

### DIFF
--- a/changelogs/fragments/220-routing-id-bgp-connection.yml
+++ b/changelogs/fragments/220-routing-id-bgp-connection.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - api_modify, api_info - support API paths ``routing id``, ``routing bgp connection`` (https://github.com/ansible-collections/community.routeros/pull/220)
+  - api_modify, api_info - support API paths ``routing id``, ``routing bgp connection`` (https://github.com/ansible-collections/community.routeros/pull/220).
--- a/changelogs/fragments/220-routing-id-bgp-connection.yml
+++ b/changelogs/fragments/220-routing-id-bgp-connection.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - api_modify, api_info - support API paths ``routing id``, ``routing bgp connection`` (https://github.com/ansible-collections/community.routeros/pull/220)
--- a/plugins/module_utils/_api_data.py
+++ b/plugins/module_utils/_api_data.py
@@ -2840,6 +2840,18 @@ PATHS = {
             },
         ),
     ),
+    ('routing', 'id'): APIData(
+        unversioned=VersionedAPIData(
+            fully_understood=True,
+            fields={
+                'disabled': KeyInfo(default=False),
+                'id': KeyInfo(),
+                'name': KeyInfo(),
+                'select-dynamic-id': KeyInfo(),
+                'select-from-vrf': KeyInfo(),
+            },
+        ),
+    ),
     ('routing', 'bfd', 'interface'): APIData(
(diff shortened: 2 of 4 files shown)


ialshaev:
Recently, I was thinking to add more BGP configuration paths as per now there is only `instance` path added, but `connection`, `networks` and `templates` are at high requirement.

I was quite surprised when I found that the configuration for `routing bgp templates` and `routing bgp connections` with completely same properties may be duplicated. It seems to be unexpected and I am curious may we handle such behaviour in the code or do we have to reporting the ROS developers?

Not sure but at my point of view it looks like a bug.

felixfontein:
> I was quite surprised when I found that the configuration for `routing bgp templates` and `routing bgp connections` with completely same properties may be duplicated. It seems to be unexpected and I am curious may we handle such behaviour in the code or do we have to report the ROS developers?
> 
> Not sure but at my point of view it looks like a bug.

I would probably ask in the ROS forums, to figure out whether it is a bug or intended behavior (and if it's intended: what's the reason for it?).

changelogs/fragments/220-routing-id-bgp-connection.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - api_modify, api_info - support API paths ``routing id``, ``routing bgp connection`` (https://github.com/ansible-collections/community.routeros/pull/220)
```
felixfontein, line 2:
```suggestion
  - api_modify, api_info - support API paths ``routing id``, ``routing bgp connection`` (https://github.com/ansible-collections/community.routeros/pull/220).
```

felixfontein (approved):

felixfontein:
@cosandr thanks for your contribution!

PR #192, opened 2023-07-21
Fixes #191: Add path 'ip traffic-flow target'

derdeagle:
##### SUMMARY
Adding the path **ip traffic-flow target** with the default values I could find.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
- api_info
- api_modify

##### ADDITIONAL INFORMATION
None.

PR #234, opened 2023-11-11
Fix some bugs and adapt RouterOS v.7.12 changes

derdeagle:
/ip traffic-flow target: make it a multiple value (single_value=False) field
/routing id: add comment attribute
/tool e-mail: add versioning to the parameters as "address" was renamed to "server" in 7.12 /routing bgp connection: add missing attributes

##### SUMMARY
/ip traffic-flow target: make it a multiple value (single_value=False) field
/routing id: add comment attribute
/tool e-mail: add versioning to the parameters as "address" was renamed to "server" in 7.12 /routing bgp connection: add missing attributes

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
api_data


plugins/module_utils/_api_data.py
```diff
@@ -2390,7 +2390,7 @@ def join_path(path):
     ),
     ('ip', 'traffic-flow', 'target'): APIData(
         unversioned=VersionedAPIData(
-            single_value=True,
+            single_value=False,
```
felixfontein, line 2393: Is this something that was already wrong in the past? If not, you probably also need to make this dependent on the ROS version.

plugins/module_utils/_api_data.py
```diff
@@ -2892,6 +2892,7 @@ def join_path(path):
         unversioned=VersionedAPIData(
             fully_understood=True,
             fields={
+                'comment': KeyInfo(can_disable=True, remove_value=''),
```
felixfontein, line 2895: Is this ROS 7.12+ only as well, or does it also apply to older versions?

changelogs/fragments/234-bugfixes-and-update-adaptations.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - api_modify - add ``comment`` attribute to ``/routing id``, make ``/ip traffic-flow target`` a multiple value attribute, add versioning to the ``/tool e-mail`` path (RouterOS 7.12 release), and add missing attributes to the ``routing bgp connection`` path (https://github.com/ansible-collections/community.routeros/pull/234).
```
felixfontein, line 2: It might make sense to split this up into multiple fragments, like this:
```suggestion
  - api_modify - add ``comment`` attribute to ``/routing id``, make ``/ip traffic-flow target`` a multiple value attribute, add versioning to the ``/tool e-mail`` path (RouterOS 7.12 release) (https://github.com/ansible-collections/community.routeros/pull/234).
  - api_modify - add missing attributes to the ``routing bgp connection`` path (https://github.com/ansible-collections/community.routeros/pull/234).
```
Potentially even into more, depending on whether the "RouterOS 7.12 release" applies to all the things mentioned before it, or only to the last.

felixfontein (commented):
Thanks for your contribution! Here are some very first comments :)

plugins/module_utils/_api_data.py
derdeagle: This was wrong in the past already. I myself added this error in PR #192.

derdeagle:
@felixfontein Thank you very much for your review. I addressed or at least answered on your comments.

felixfontein (approved):

felixfontein:
@derdeagle thanks for your improvements!

PR #266, opened 2024-03-04
Add interface wifi paths

### LINKED ISSUE #253: Add support for newer 'wifi' interface
If possible please also add support for the `wifi` interface type. Config is similar to `wifiwave2`, after 7.12 it replaces the `wifiwave2` package actually.

Thank you

cosandr:
##### SUMMARY
Fixes #253.

Mostly copy-pasted from `interface wifiwave2` with some minor changes (added missing `comment` and `disabled` where relevant, removed `openflow-switch` from `datapath`, changed default l2mtu to 1560).

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
* `api_info`
* `api_modify`

##### ADDITIONAL INFORMATION

I'm unsure about the default values, but they haven't caused me any issues.

I'm unable to rename the default wifi interfaces, it fails with
> Error while removing name=\"wifi1\" (ID *6): failure: not allowed to remove

I can do it from the CLI with
```
/interface/wifi> set wifi1 name=wifi-test
```

I assume the issue is having `name` as primary key, causing the module to try to remove it before creating a new one with the new name.


plugins/module_utils/_api_data.py
```diff
@@ -1533,6 +1533,303 @@ def join_path(path):
...
         ),
     ),
+    ('interface', 'wifi'): APIData(
+        versioned=[
+            ('7.13', '>=', VersionedAPIData(
+                fully_understood=True,
+                primary_keys=('name', ),
+                required_one_of=[['default-name', 'master-interface']],
+                fields={
+                    'aaa': KeyInfo(),
```
derdeagle, line 1543: The `aaa` propery is for referencing a setting under `interface wifi aaa` but one can also define the values like `aaa.called-format`, `aaa.calling-format`, `aaa.interim-update`, etc. As you set the `fully_understood` flag to `True` (which I highly encourage) would you mind adding all possible settings for `aaa.*`, `channel.*`, `configuration.*`, `datapath.*`, `interworking.*`, `security.*`, and `steering.*`?

derdeagle: Also for at least 7.14 I get the following error when executing `/interface/wifi/add` without any argument:
`failure: must specify exactly one of radio-mac and master-interface`

I cannot see `default-name` on my device.
In the release notes for version 7.13 I found the following:

`*) wifi - added "radio-mac" variable for "name-format" provisioning setting;`

plugins/module_utils/_api_data.py
```diff
@@ -1533,6 +1533,303 @@ def join_path(path):
             },
         ),
     ),
+    ('interface', 'wifi'): APIData(
```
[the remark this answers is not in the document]
cosandr: You're right, I missed it. I'll add it.

plugins/module_utils/_api_data.py
cosandr: OK, I've added them.

cosandr: Yep, I get the same error on 7.13. I've removed `default-name`, I don't see it so I assume it's gone.

cosandr: FYI I've added back `default-name`, it's a name-only property but I think it makes sense to keep it there as it makes configuring the interfaces easier (since you don't need to always use `radio-mac` for the masters).

changelogs/fragments/266-interface-wifi.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - api_info, api_modify - added support for ``interface wifi`` (https://github.com/ansible-collections/community.routeros/pull/266).
```
felixfontein, line 2:
```suggestion
  - api_info, api_modify - added support for ``interface wifi`` and its sub-paths (https://github.com/ansible-collections/community.routeros/pull/266).
```

felixfontein:
The module only knows how to remove values that have `can_disabled=True` (optionally with `remove_value`, depending on how to remove it via the API). Since most of the fields do not have `can_disabled=True` the module doesn't know how to remove them, and since you supplied `handle_entries_content: remove_as_much_as_possible` it does not complain in that case.

cosandr:
I see, thanks! I've tried a few more things and most paths only require the primary key, everything else can be disabled, so I've removed most defaults in favor of `can_disable=True`. Some have to stay (`interface wifi` needs at least `l2mtu` and `arp-timeout`). I can't guarantee every single option works, but I think not being able to remove them at all is not expected behavior.

I'm also thinking it's better to remove them instead of duplicating rOS defaults. For example even though the docs claim `arp=enabled` is the default, it doesn't show up if you run `export` whereas it would if the module removes it and sets it back to the default value, whereas it won't show up if it's removed via `can_disable=True` (so it's the same as it was before the module did anything with it).

felixfontein:
I don't have time to properly review this, but from a glance it looks good. If nobody objecs, I'll merge this tomorrow :)

PR #288, opened 2024-05-16
minor changes '/interface ethernet' path fields

vint2k:
##### SUMMARY
- added remove_value for `fec-mode`, `poe-out` and `poe-priority` fields because if these fields undefined it will try to remove default value on sfp and poe ports
- fixed default value for `sfp-shutdown-temperature` field - 95 instead 95C (95C always will show diffs)
- removed default value for `l2mtu` field because different hardware has own default values, ex. RB4011 `l2mtu` is **1592**, but CCR2116 `l2mtu` is **1584** and `l2mtu` of sfp ports on both is **1600**

##### ISSUE TYPE
- Bugfix Pull Request


felixfontein (approved):
Looks good to me!

felixfontein:
@vint2k thanks for your contribution!

PR #300, opened 2024-07-21
Add IPv4 DHCP option matcher, minor fixes

hansmi:
##### SUMMARY

RouterOS 7.4 and newer implement matchers for the Vendor ID provided by DHCP clients under `ip dhcp-server matcher`. The default value for `force` on options is also set.

##### ISSUE TYPE
- Feature Pull Request


changelogs/fragments/300-set-ip-dhcp-option-force-default.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - api_info, api_modify - set default for ``force`` in ``ip dhcp-server option`` to an explicit false (https://github.com/ansible-collections/community.routeros/issues/300).
```
felixfontein, line 2: BTW, you can also put both fragments into the same YML file, as two list entries in `minor_changes`.

```suggestion
  - api_info, api_modify - set default for ``force`` in ``ip dhcp-server option`` to an explicit ``false`` (https://github.com/ansible-collections/community.routeros/issues/300).
```

felixfontein (commented):
Thanks for your contribution!

changelogs/fragments/300-set-ip-dhcp-option-force-default.yml
hansmi: Done. I deliberately used separate files because they're added by separate commits.

felixfontein: Fine for me. But please note that the commits will get squashed on merge anyway :)

felixfontein (approved):

felixfontein:
@hansmi thanks!

PR #301, opened 2024-07-27
[WIP] Try to not install python-daemon 3.0.2 in EEs

felixfontein:
##### SUMMARY
Work around https://pagure.io/python-daemon/issue/94.

##### ISSUE TYPE
- Test Pull Request

##### COMPONENT NAME
EE tests


felixfontein:
To be honest the current solution (which works) looks like a *huge* hack. Having a proper way to pin / constraint dependencies would be better...

Ref: https://github.com/ansible/ansible-builder/issues/699

felixfontein:
python-daemon 3.0.2 has been yanked.

PR #305, opened 2024-08-11
api_modify/api_info: add `restrict` option

felixfontein:
##### SUMMARY
Allows to restrict returned information / modified data by forcing fields to specific values. For example, for `ip firewall filter` you can limit `chain` to `input` to only query / modify the `input` chain, without touching the other chains.

Closes #160.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
api_info
api_modify


plugins/module_utils/_api_helper.py
```diff
@@ -0,0 +1,40 @@
...
+        return True
+    for rule in restrict:
+        field = rule['field']
+        field_info = path_info.fields[field]
+        value = entry.get(field)
+        if value is None:
+            value = field_info.default
+        if field not in entry and field_info.absent_value:
+            value = field_info.absent_value
+        if value not in rule['values']:
```
hansmi, line 38: Could we also support regular expressions?

As you may know RouterOS does not (yet) support automatic DNS records for DHCP leases. There are numerous script approaches out there and it's useful to track the corresponding lease in a comment. If mixed with static entries managed via Ansible and `handle_absent_entries: remove` those added by a script would be removed. By specifying a regex rejecting the script-managed records my current approach would be simpler:

1. Request all DNS static records (optionally filtered by having a comment), extract only those managed by scripts
2. Combine with records from Ansible variables
3. Re-apply

felixfontein: Sure, that should be no problem!

felixfontein: 2baeb174e375e753d7f69b7d79413ad8cc45d390

hansmi: Thank you!

plugins/doc_fragments/api.py
```diff
@@ -95,3 +95,35 @@ class ModuleDocFragment(object):
...
+            both the string and the integer.
+          - Use V(none) for disabled values.
+          - Either O(restrict[].values) or O(restrict[].regex), but not both, must be specified.
+        type: list
+        elements: raw
+      regex:
+        description:
+          - A regular expression matching values of the field to limit to.
+          - Note that all values will be converted to strings before matching.
+          - It is not possible to match disabled values with regular expressions.
```
hansmi, line 126: With an additional option, e.g. `negative`, matching disabled/non-existent values should be possible. Consider the following config:

```
/ip dns static
add address=1.2.3.4 comment="# managed, lease=foobar123" disabled=no name=test.example.com ttl=1d
add address=2.3.4.5 disabled=no name=static.example.com ttl=1d
```

If I understand the logic correctly it's not possible to restrict the module's operations to the entries _without_ comment.

felixfontein: While inversion/negation makes sense, it is possible to restrict to everything without comments by specifying:
```
  - field: ...
    values:
      - ~
```

hansmi: You're right. My regex-colored glasses made me miss `values: [null]`. That considered it might be more flexible to support both matchers at the same time:

```yaml
restrict:
- value: "hello"
- value: null
- regex: "^foo"
- value: "world"
  exclude: true
- regex: "^# magic"
  exclude: true
```

felixfontein: I changed the behavior in 6e59f36a1b3871b6262d5c8638faa4547f758b67, now you have two additional options `match_disabled` and `invert`, and `values` and `regex` are no longer mutually exclusive. So you can do:
```yaml
restrict:
  # Match all entries which do not have a comment
  - field: comment
    match_disabled: true
    invert: true
  # Match chain "input" and all chains starting with "foo"
  - field: chain
    values:
      - input
    regex: "^foo"
```

hansmi: LGTM, thank you a lot!

felixfontein:
This is now ready :) I'll merge this later today and create a new release (for inclusion in ~tomorrow's Ansible 9/10 releases).

felixfontein:
@hansmi thanks for reviewing and commenting!

PR #313, opened 2024-09-07
Added support for 'routing filter num-list'

tim427:
##### SUMMARY
Added `routing filter num-list` support, similar as `routing filter rules`

##### ISSUE TYPE
- New Module Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/module_utils/_api_data.py
+++ b/plugins/module_utils/_api_data.py
@@ -844,6 +844,19 @@ PATHS = {
             )),
         ],
     ),
+    ('routing', 'filter', 'num-list'): APIData(
+        versioned=[
+            ('7', '>=', VersionedAPIData(
+                fully_understood=True,
+                fields={
+                    'list': KeyInfo(required=True),
+                    'comment': KeyInfo(can_disable=True, remove_value=''),
+                    'disabled': KeyInfo(can_disable=True),
+                    'range': KeyInfo(can_disable=True),
+                },
+            )),
+        ],
+    ),
     ('routing', 'filter', 'rule'): APIData(
         versioned=[
             ('7', '>=', VersionedAPIData(
--- a/changelogs/fragments/313-add-routing-filter-num-list.yml
+++ b/changelogs/fragments/313-add-routing-filter-num-list.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - api_info, api_modify - add support for the ``routing filter num-list`` path implemented by RouterOS 7.1 and newer (https://github.com/ansible-collections/community.routeros/pull/313).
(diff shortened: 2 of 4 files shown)


plugins/module_utils/_api_data.py
```diff
@@ -844,6 +844,19 @@ def join_path(path):
             )),
         ],
     ),
+    ('routing', 'filter', 'num-list'): APIData(
+        versioned=[
+            ('7', '>=', VersionedAPIData(
+                fully_understood=True,
```
felixfontein, line 850: Should this have a primary key? I think `("list", "range")` might make sense - or even `("list", )`? (I don't have access to a device right now so I cannot try this out.)

felixfontein (commented):
Thanks for your contribution! Could you please add a [changelog fragment](https://docs.ansible.com/ansible/devel/community/collection_development_process.html#creating-a-changelog-fragment)? Thanks.

plugins/module_utils/_api_data.py
tim427: Yeah, I was fiddling around with the `primary_keys=('list', 'range')` and `primary_keys=('list', )`, but I don't see any difference in the output.

Or at least, I doesn't change the order, all though I've `ensure_order: true`, `handle_absent_entries: 'remove'` and `handle_entries_content: 'remove'`.

So I'm a little bit clueless, at this moment.

felixfontein: The question is whether you can have more than one entry with the same `list` or not. If not: `primary_keys=('list', )`. Otherwise, if you can have only one entry with the same `list` and `range`, then it's `primary_keys=('list', 'range')`. If not, then it's something else...

tim427: It's the latter. You could have multiple `range` on the same `list`:

image

felixfontein: Cool! But you cannot have multiple entries with the same `list` and `range`, right?

plugins/module_utils/_api_data.py
```diff
@@ -844,6 +844,19 @@ def join_path(path):
             )),
         ],
     ),
+    ('routing', 'filter', 'num-list'): APIData(
+        versioned=[
+            ('7', '>=', VersionedAPIData(
```
felixfontein, line 849: Since this is only supported since RouterOS 7.1 (according to the changelog fragment you added), this should be:
```suggestion
            ('7.1', '>=', VersionedAPIData(
```

tim427: As far as I know, this is there from the beginning, and the first version is 7.1  (according to the changelogs of MikroTik). So yeah :) But the same holds for the other functions.

I you want, I could change the changelog to `7` instead of `7.1`.

plugins/module_utils/_api_data.py
tim427: image
Apparently, that's possible :)

So, no "uniques", and therefore no "primary" at all? Besides the `id`.

plugins/module_utils/_api_data.py
felixfontein: Maybe change the changelog to 7 then.

plugins/module_utils/_api_data.py
felixfontein: Hmm, that's weird :) Yeah, in that case no primary key is probably best...

felixfontein (approved):

felixfontein:
@tim427 thanks again for your contribution!

PR #324, opened 2024-10-23
Update _api_data.py

liquorice-head:
##### SUMMARY

The behavior of the log and log-prefix fields has been changed. Previously, if these fields were not declared in the code, Ansible would attempt to delete them. Now, if the fields are not declared, Ansible will assign them the default value for the router OS

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

- _api_data.py

##### ADDITIONAL INFORMATION

```paste below

```


plugins/module_utils/_api_data.py
```diff
@@ -2984,8 +2984,8 @@ def join_path(path):
                 'jump-target': KeyInfo(can_disable=True),
                 'layer7-protocol': KeyInfo(can_disable=True),
                 'limit': KeyInfo(can_disable=True),
-                'log': KeyInfo(can_disable=True),
-                'log-prefix': KeyInfo(can_disable=True),
+                'log': KeyInfo(default=False),
+                'log-prefix': KeyInfo(default=''),
```
felixfontein, line 2988: What's the reason behind this change? Was the behavior always wrong, or did RouterOS's behavior change from one version to another?

felixfontein (commented):
Thanks for your contribution!

plugins/module_utils/_api_data.py
liquorice-head: image
As far as I'm concerned, the behavior was always wrong. I have tested on 6.49.10, 6.49.13, 7.14.1. If you use the --diff parameter in asible, you will see that each time ansible tries to remove these fields.

changelogs/fragments/324-fix-firewall-log-and-log-prefix.yaml
```diff
@@ -0,0 +1,2 @@
+bugfixes:
+   - fields 'log' and 'log-prefix' in paths ip firewall filter, mangle, nat, raw changed from KeyInfo(can_disable=True) to KeyInfo(default=False) for 'log' and KeyInfo(default='') for 'log-prefix' (https://github.com/ansible-collections/community.routeros/pull/324).
```
felixfontein, line 2:
```suggestion
   - api_info, api_modify - fields ``log`` and ``log-prefix`` in paths ``ip firewall filter``, ``ip firewall mangle``, ``ip firewall nat``, ``ip firewall raw`` now have the correct default values (https://github.com/ansible-collections/community.routeros/pull/324).
```

felixfontein (commented):
Please note that the unit tests now fail, since some of them rely on the old structure of I think `ip firewall filter`.

PR #327, opened 2024-11-01
fix: add missing fields to `ip pool` path

gardar:
##### SUMMARY

add missing fields `comment`, `copy-from`, `next-pool` to `ip pool` path.
Not sure in what version if routeros they were introduced, the oldest I have access to is 6.46 and they are present there.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

`api_info`
`api_modify`
`ip pool`

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/module_utils/_api_data.py
+++ b/plugins/module_utils/_api_data.py
@@ -709,7 +709,6 @@ PATHS = {
             primary_keys=('name', ),
             fields={
                 'comment': KeyInfo(),
-                'copy-from': KeyInfo(),
                 'name': KeyInfo(),
                 'next-pool': KeyInfo(),
                 'ranges': KeyInfo(),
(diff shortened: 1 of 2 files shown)


plugins/module_utils/_api_data.py
```diff
@@ -708,7 +708,10 @@ def join_path(path):
             fully_understood=True,
             primary_keys=('name', ),
             fields={
+                'comment': KeyInfo(),
+                'copy-from': KeyInfo(),
```
felixfontein, line 712: Isn't `copy-from` a parameter for creation only, and not a propery of the actual pools?

felixfontein (commented):
Thanks for your contribution!

plugins/module_utils/_api_data.py
gardar: Yes good point. Do I mark it for creation only somehow or should I remove it?

felixfontein: I would remove it. `copy-from` is a parameter to make creation of new entries easier from the command line.

felixfontein (approved):

felixfontein:
@gardar thank you!

PR #341, opened 2025-01-26
DHCPv6 client: add missing 'use-interface-duid' option

petrows:
##### SUMMARY
api_info, api_modify: add the ``use-interface-duid`` option for DHCPv6 client. 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
api_info, api_modify

##### ADDITIONAL INFORMATION
This option prevents issues with Fritzbox modems and routers, when using virtual interfaces (like VLANs) may create duplicated records in hosts config, this breaks original "expose-host" function.

The ``use-interface-duid`` corrects this behavior: the DUID for link-local address and DHCPv6 request are the same, and user can configure per-host option in Fritzbox OS as expected.

Tested on current ``7.17`` and ``7.16.2`` versions:

```diff
--- before
+++ after
@@ -12,6 +12,7 @@
             "pool-prefix-length": 64,
             "prefix-hint": "::/59",
             "request": "address,prefix",
+            "use-interface-duid": true,
             "use-peer-dns": false
         }
     ]
```

### DIFF
--- a/changelogs/fragments/341-add-dhcpv6-client-use-interface-duid.yml
+++ b/changelogs/fragments/341-add-dhcpv6-client-use-interface-duid.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - api_info, api_modify - add the ``use-interface-duid`` option for ``ipv6 dhcp-client`` path. This option prevents issues with Fritzbox modems and routers, when using virtual interfaces (like VLANs) may create duplicated records in hosts config, this breaks original "expose-host" function. Also add the ``script``, ``custom-duid`` and ``validate-server-duid`` as backport from 7.15 version update (https://github.com/ansible-collections/community.routeros/pull/341).
--- a/plugins/module_utils/_api_data.py
+++ b/plugins/module_utils/_api_data.py
@@ -3544,6 +3544,14 @@ PATHS = {
                 'request': KeyInfo(),
                 'use-peer-dns': KeyInfo(default=True),
             },
+            versioned_fields=[
+                # Mikrotik does not provide exact version in official changelogs.
+                # The 7.15 version is the earliest, found option in router config backups:
+                ([('7.15', '>=')], 'script', KeyInfo(default='')),
+                ([('7.15', '>=')], 'custom-duid', KeyInfo(default='')),
+                ([('7.15', '>=')], 'use-interface-duid', KeyInfo(default=False)),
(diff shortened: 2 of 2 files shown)


changelogs/fragments/341-add-dhcpv6-client-use-interface-duid.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - api_info, api_modify - add the ``use-interface-duid`` option for DHCPv6 client. This option prevents issues with Fritzbox modems and routers, when using virtual interfaces (like VLANs) may create duplicated records in hosts config, this breaks original "expose-host" function.
```
felixfontein, line 2:
```suggestion
  - api_info, api_modify - add the ``use-interface-duid`` option for ``ipv6 dhcp-client`` path. This option prevents issues with Fritzbox modems and routers, when using virtual interfaces (like VLANs) may create duplicated records in hosts config, this breaks original "expose-host" function (https://github.com/ansible-collections/community.routeros/pull/341).
```

plugins/module_utils/_api_data.py
```diff
@@ -3544,6 +3544,11 @@ def join_path(path):
                 'request': KeyInfo(),
                 'use-peer-dns': KeyInfo(default=True),
             },
+            versioned_fields=[
+                # Mikrotik does not provide exact version in official changelogs.
+                # The 7.15 version is the earliest, found option in router config backups:
+                ([('7.15', '>=')], 'use-interface-duid', KeyInfo(default=False)),
```
felixfontein, line 3550: How about also adding the other DUID related fields? `custom-duid` (no default) and `validate-server-duid` (default: `True`). I don't know when these options were added, but I guess adding 7.15 here as well makes sense.

felixfontein (commented):
Thanks for your contribution!

plugins/module_utils/_api_data.py
petrows: Thanks for the hint, @felixfontein ! I have added missing options, also `script` was missing as well, also added.

changelogs/fragments/341-add-dhcpv6-client-use-interface-duid.yml
petrows: Requested changes applied, also added list of back-ported functions from 7.15 (?) version.

felixfontein (approved):

felixfontein:
@petrows thanks again for your contribution!

PR #380, opened 2025-07-22
Update IPv6 settings fields for RouterOS 7.17/7.18

hansmi:
##### SUMMARY

From the RouterOS 7.17 changelog:

*) ipv6 - added IPv6 settings related to stale IPv6 neighbor cleanup; *) ipv6 - added support for manual link-local address configuration;

And from the RouterOS 7.18 changelog:

*) ipv6 - added FastTrack support;
*) ipv6 - added support for neighbor removal and static entries;

The change also removes the default value for the `max-neighbor-entries`
field. It's documented to vary depending on the RAM size.


felixfontein:
@hansmi thank you!

PR #392, opened 2025-08-10
Add show-at-cli-login property for system note

hansmi:
##### SUMMARY

RouterOS 7.14 (March 2024) added `show-at-cli-login` in `system note`.

##### ISSUE TYPE
- Feature Pull Request


felixfontein (approved):

plugins/module_utils/_api_data.py
```diff
@@ -4423,6 +4423,7 @@ def join_path(path):
             fully_understood=True,
```
felixfontein, line 4423: Maybe this way would be better though:
```suggestion
            fully_understood=True,
            versioned_fields=[
                ([('7.14', '>=')], 'show-at-cli-login', KeyInfo(default=False)),
            ],
```

felixfontein (approved):

PR #394, opened 2025-08-12
Set default value for interface list include/exclude

hansmi:
##### SUMMARY
Without a default value a pre-existing value isn't removed unless the caller specifies the property.

##### ISSUE TYPE
- Bugfix Pull Request


felixfontein (approved):

felixfontein:
@hansmi thanks!

PR #420, opened 2025-11-28
Solves issue #409: remove fixed_entries=True for path /routing/bgp/in…

### LINKED ISSUE #409: Cannot add new entry to this path: routing bgp instance
##### SUMMARY
The new `routing bgp instance` (routeros 7.20) path does not work.
I only get "Cannot add new entry to this path" error.

Like in #256 and because there is no predefined bgp instance, I guess that `fixed_entries=True` must be removed in `_api_data.py`.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
`routing bgp instance`

##### COLLECTION VERSION
3.12.1

##### HOW TO REPRODUCE
You can reproduce the problem with this simple data.
```yaml
  - { name: bgp-instance-test, as: 65123, router-id: 192.168.1.254 }
```

SdVSysAdmin:
##### SUMMARY

Solves issue #409: remove fixed_entries=True for path `/routing/bgp/instance`

Because there is no predefined bgp instance, I guess that `fixed_entries=True` must be removed in `_api_data.py`.

Fixes #409.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

path `/routing/bgp/instance`

### DIFF
--- a/changelogs/fragments/420_fix_routing_bgp_instance.yml
+++ b/changelogs/fragments/420_fix_routing_bgp_instance.yml
@@ -1,3 +1,3 @@
 ---
 bugfixes:
-  - api_modify, api_info - in the ``routing bgp instance``, fix 'Cannot add new entry to this path' error (https://github.com/ansible-collections/community.routeros/issues/409, https://github.com/ansible-collections/community.routeros/pull/420).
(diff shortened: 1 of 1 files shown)


felixfontein:
@SdVSysAdmin thanks for your contribution!

PR #421, opened 2025-11-28
Modify support for path ip/ipsec/identity

SdVSysAdmin:
##### SUMMARY

The current `ip/ipsec/identity` defines `peer` as a primary key which seems wrong for me. On routeros it is possible to use the same peer config in several identity. 

Removing the `primary_keys` constraint enable the definition of several user with the same psk and peer definition.

I have not found any other potential primary key in the `/ip/ipsec/identity` parameters.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

 `ip/ipsec/identity`  path.


felixfontein (approved):
Thanks for fixing this!

Issue #363, opened 2025-04-20
/system/note set show-at-cli-login


ivoruetsche:
##### SUMMARY

Just a cosmetic parameter, but would be nice to have it:
/system/note set show-at-cli-login
https://help.mikrotik.com/docs/spaces/ROS/pages/40992863/Note

##### ISSUE TYPE
- Implementing /system/note set show-at-cli-login

##### ADDITIONAL INFORMATION

```yaml

```

ivoruetsche:
sorry, I didn't got the latest release. Is implemented thanks a lot

ivoruetsche:
I was wrong, it's not

ivoruetsche:
Some more in `ip neighbor discovery-settings`:
- `lldp-max-frame-size`
- `lldp-poe-power`
- `lldp-dcbx`

Great thanks a lot

PR #423, opened 2025-12-02
Partially fixed issue 363: Add missing parameters for path interface bridge and interface bridge port

FelixRicke:
##### SUMMARY
Add missing parameters to path interface bridge and interface bridge port

Partially fixes issue #363.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
- _api_data.py

##### ADDITIONAL INFORMATION

### DIFF
--- a/changelogs/fragments/423_add_missing_parameters_for_interface_bridge.yml
+++ b/changelogs/fragments/423_add_missing_parameters_for_interface_bridge.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - api_info, api_modify - add missing parameters to path ``interface bridge`` and ``interface bridge port`` (https://github.com/ansible-collections/community.routeros/pull/423)
+  - api_info, api_modify - add missing parameters to path ``interface bridge`` and ``interface bridge port`` (https://github.com/ansible-collections/community.routeros/pull/423).


FelixRicke:
Hi @felixfontein,

I am not currently working with older versions of RouterOS (nor am I familiar with them),
so I cannot say when a path was assigned new parameters or default values.
And to be honest, I don't really have the time to check every version or every release note (which, in my opinion, often do not provide direct information).
So what do you think would be the best approach? Should I always add the changes as versioned_fields?

felixfontein:
> I am not currently working with older versions of RouterOS (nor am I familiar with them), so I cannot say when a path was assigned new parameters or default values. And to be honest, I don't really have the time to check every version or every release note (which, in my opinion, often do not provide direct information). So what do you think would be the best approach? Should I always add the changes as versioned_fields?

That's a very good question, and something I've been wondering about as well. The RouterOS documentation is really bad about such things... I wish there would be some (automatically generated / machine readable) API docs that could be used as a basis for the API modules in this collection, but there don't seem to be any.

Simply adding the new fields with the "current" minor version - as you did in this PR - sounds like a good idea; if someone is using an older version and notices that something is different there, they can create a bugfix to fix the version numbers. If nobody is using them, well, then it doesn't matter anyway :)

felixfontein (approved):

PR #427, opened 2025-12-16
Vrrp group authority

SdVSysAdmin:
##### SUMMARY

In path `interface/vrrp`, `group-master` is renamed to `group-authority` in routeros 7.11. (https://forum.mikrotik.com/t/v7-11-2-stable-is-released/168778).

##### ISSUE TYPE

`group-master` is still present in lastest routeros release but only for backward compatibility. So `group-master` works but the task is not idempotent. It always have `changed` status. Using `group-authority` restores the idempotent property.

##### COMPONENT NAME

`interface/vrrp`

##### ADDITIONAL INFORMATION

This PR can break previously working tasks/playbooks as it require to replace `group-master` by `group-authority` if running routeros 7.11+. For me this is acceptable as idempotent is an important feature of ansible.

### DIFF
--- a/changelogs/fragments/427_vrrp_group_authority.yml
+++ b/changelogs/fragments/427_vrrp_group_authority.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - api_info, api_modify - rename group-master to group-authority in interface/vrrp on routeros 7.11+


felixfontein (commented):
Thanks!

felixfontein:
Ping @SdVSysAdmin

PR #430, opened 2026-01-06
Add prefix-pool field to ipv6 dhcp-server path

vfreex:
Add missing prefix-pool parameter to the IPv6 DHCP server configuration path. Both address-pool and prefix-pool now default to 'static-only' instead of being required, matching the IPv4 DHCP server pattern.


vfreex:
Hi @felixfontein, I've pushed an update with the changelog.
> Do you happen to know since which RouterOS version this has changed? It would be great to make this version-dependent.
I didn't find any official docs mentioning this. I've been using prefix-pool to assign ipv6 prefixes to downstream routers for more than 2 years, so I guess it is not a new feature. Is it required to make it version-dependent? Does this project have any guidelines/policies about minimum routeros version to support?

PR #439, opened 2026-02-07
add defaults for fields mvrp, mvrp-forbidden, ciphers

pypb:
##### SUMMARY

When testing v3.17.0-a1, ansible-playbook failed on api_modify for paths interface/vlan, interface/bridge and ip/ssh with *Key cannot be removed*.

This is due to new fields in _api_data.py without defaults values set. I've added the defaults described in the RouterOS documentation to these fields.

We should probably go through all the fields at some point, though...

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

_api_data.py

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/changelogs/fragments/439-add-defaults-for-fields.yml
+++ b/changelogs/fragments/439-add-defaults-for-fields.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - _api_data.py - add default values for fields ``ip/ssh``:``ciphers``, ``interface/vlan``:``mvrp``, ``interface/bridge``:``mvrp-forbidden``.
+  - "api_info, api_modify - add default values for fields ``ciphers`` in ``ip ssh``, ``mvrp`` in  ``interface vlan``, and ``mvrp-forbidden`` in ``interface bridge vlan`` (https://github.com/ansible-collections/community.routeros/pull/439/)."


changelogs/fragments/439-add-defaults-for-fields.yml
```diff
@@ -0,0 +1,2 @@
+minor_changes:
+  - _api_data.py - add default values for fields ``ip/ssh``:``ciphers``, ``interface/vlan``:``mvrp``, ``interface/bridge``:``mvrp-forbidden``.
```
felixfontein:
```suggestion
  - "api_info, api_modify - add default values for fields ``ciphers`` in ``ip ssh``, ``mvrp`` in  ``interface vlan``, and ``mvrp-forbidden`` in ``interface bridge vlan`` (https://github.com/ansible-collections/community.routeros/pull/439/)."
```
This PR does not contain any changes for `mvrp` in `interface vlan` though.

pypb: Oups! Good catch, fixed.

felixfontein:
I think for now it's good to get more testing (as always ;) ) and make sure that all reported problems are fixed. I'll likely make another 2.17.0-a2 release tomorrow with this PR in it (and more if something shows up until then), if that's OK with everyone.

PR #445, opened 2026-02-13
add `primary_keys=('name',)` to path `container` for RouterOS >= 7.19

### LINKED ISSUE #443: api_modify container - not possible to modify existing container
Related with [v3170-a1 release](https://github.com/ansible-collections/community.routeros/blob/main/CHANGELOG.md#v3170-a1).

Is it possible to add the new container, for isntance like this:
```yaml
    - name: Container add
      community.routeros.api_modify:
        path: container
        data:
          - name: nginx
            remote-image: nginx
            interface: veth1
            ...
            entrypoint: "sleep"
            cmd: "infinity"
            hostname: my.hostname
```

But, afterwards it is not possible to modify this created container (in case if I change some parameter and re-run the playbook), there is such routeros built-in error:
```log
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Error while creating entry: failure: Name already exists"}
```

I assume in this case we should use `container update ...` instead of `container add ...` 

Versions:
```
Ansible v2.20.2
librouteros 3.4.2
ansible.netcommon 8.2.1
community.routeros 3.17.0-a2
```

Env:
```
Mikrotik RouterOS 7.21
```

FelixRicke:
##### SUMMARY
Adds `primary_keys=('name',)` to the `container` path for RouterOS versions 7.19 and later.

Fixes #443 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
- _api_data.py

##### ADDITIONAL INFORMATION
Before RouterOS 7.19, the `name` parameter was a read-only field that was automatically generated when creating a container.
Starting with version 7.19, it is possible to assign and modify a container’s name manually.
Although RouterOS can still auto-generate this parameter, it must remain unique.
To ensure correct handling when modifying an existing container, `name` is now declared as a `primary_key` in the `api_modify` module for RouterOS versions `>= 7.19`.

### DIFF
--- a/changelogs/fragments/445-fix-container-paths-name-parameter.yml
+++ b/changelogs/fragments/445-fix-container-paths-name-parameter.yml
@@ -1,2 +1,2 @@
 minor_changes:
-  - api_info, api_modify - add ``primary_keys=('name',)`` to the ``container`` path for RouterOS >=7.19 (https://github.com/ansible-collections/community.routeros/issues/443, https://github.com/ansible-collections/community.routeros/pull/445).
+  - api_info, api_modify - make ``name`` a primary key (and thus required) for the ``container`` path for RouterOS >=7.19 (https://github.com/ansible-collections/community.routeros/issues/443, https://github.com/ansible-collections/community.routeros/pull/445).


felixfontein (commented):
Thanks!

PR #466, opened 2026-06-04
Add RouterOS 7.21 advertise-dns=self support for ipv6 nd

vfreex:
##### SUMMARY

RouterOS 7.21 added "self" as a valid value for the IPv6 ND advertise-dns setting. Model advertise-dns as a string enum on RouterOS 7.21 and later while preserving the previous boolean default for older versions.

##### ISSUE TYPE

- Feature Pull Request

##### ADDITIONAL INFORMATION

Add tests for the version boundary and idempotence with yes, no, and self values.

```paste below

```

### DIFF
--- a/plugins/module_utils/_api_data.py
+++ b/plugins/module_utils/_api_data.py
@@ -9385,10 +9385,11 @@ PATHS = {
                 ([('7.15', '>=')], 'comment', KeyInfo()),
                 # ([('7.15', '>=')], 'copy-from', KeyInfo(write_only=True)),
                 ([('7.15', '>=')], 'numbers', KeyInfo(read_only=True)),
+                ([('7.21', '<')], 'advertise-dns', KeyInfo(default=True)),
+                ([('7.21', '>=')], 'advertise-dns', KeyInfo(default='yes')),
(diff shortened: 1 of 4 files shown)


plugins/module_utils/_api_data.py
```diff
@@ -9385,10 +9385,11 @@ def join_path(path):
                 ([('7.15', '>=')], 'comment', KeyInfo()),
                 # ([('7.15', '>=')], 'copy-from', KeyInfo(write_only=True)),
                 ([('7.15', '>=')], 'numbers', KeyInfo(read_only=True)),
+                ([('7.21', '<')], 'advertise-dns', KeyInfo(default=True)),
+                ([('7.21', '>=')], 'advertise-dns', KeyInfo(default='yes')),
```
felixfontein, line 9389: I don't think this change is necessary, since librouteros automatically converts the string `"yes"` to the boolean value `True`. So the current version should already work fine.

Do you have any indication that it does not work fine?

vfreex:
Thanks for the review!  So this PR is not needed as a feature change. I’ll close it.

PR #475, opened 2026-08-26
`address` in `ip service` path is renamed to `available-from` in RouterOS 7.24

SdVSysAdmin:
##### SUMMARY

api_info, api_modify - the address parameter in the ip service path was renamed to available-from for RouterOS >= 7.24. The old name is still supported for RouterOS < 7.24.

##### COMPONENT NAME

`/ip/service` path.

##### ADDITIONAL INFORMATION

I don't known if it is possible to support both `address` and `available-from` on RouterOS >= 7.24. If yes, feel free to improve this MR.

### DIFF
--- a/changelogs/fragments/ip-service-available-from.yml
+++ b/changelogs/fragments/ip-service-available-from.yml
@@ -1,3 +1,3 @@
 minor_changes:
-  - api_info, api_modify - the ``address`` parameter in the ``ip service`` path was renamed to ``available-from`` for RouterOS >= 7.24. The old name is still supported for RouterOS < 7.24.
-    (https://github.com/ansible-collections/community.routeros/pull/475)
+  - api_info, api_modify - the ``address`` parameter in the ``ip service`` path was renamed to ``available-from`` for RouterOS >= 7.24. The old name is still supported for RouterOS < 7.24
+    (https://github.com/ansible-collections/community.routeros/pull/475).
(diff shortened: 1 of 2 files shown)


changelogs/fragments/ip-service-available-from.yml
```diff
@@ -0,0 +1,3 @@
+minor_changes:
+  - api_info, api_modify - the ``address`` parameter in the ``ip service`` path was renamed to ``available-from`` for RouterOS >= 7.24. The old name is still supported for RouterOS < 7.24.
+    (https://github.com/ansible-collections/community.routeros/pull/475)
```
felixfontein:
```suggestion
  - api_info, api_modify - the ``address`` parameter in the ``ip service`` path was renamed to ``available-from`` for RouterOS >= 7.24. The old name is still supported for RouterOS < 7.24
    (https://github.com/ansible-collections/community.routeros/pull/475).
```

felixfontein (commented):
Thanks for your contribution!

felixfontein:
> I don't known if it is possible to support both `address` and `available-from` on RouterOS >= 7.24. If yes, feel free to improve this MR.

If RouterOS still accepts `address` in 7.24+, we can add it as write-only. This won't be idempotent, though, so it's generally better to not do that.

felixfontein (approved):

felixfontein:
@SdVSysAdmin thanks for your contribution!

PR #477, opened 2026-09-07
api_info, api_modify - adds support for new parameter in the container mounts path for RouterOS

maksimsamt:


plugins/module_utils/_api_data.py
```diff
@@ -1116,14 +1116,17 @@ def join_path(path):
             ('7.22', '>=', VersionedAPIData(
                 fully_understood=True,
                 primary_keys=('dst', 'list'),
+                versioned_fields=[
+                    ([('7.23', '>=')], 'mode', KeyInfo()),
+                    ([('7.22', '>='), ('7.23', '<')], 'read-only', KeyInfo()),
```
felixfontein, line 1121: Since the whole path definition is `>= 2.22`, there's no need to repeat it here:
```suggestion
                    ([('7.23', '<')], 'read-only', KeyInfo()),
```

felixfontein (commented):
Thanks for your contribution!

maksimsamt:
@felixfontein ,
As far as I understand, the next release will not be soon?

felixfontein:
> As far as I understand, the next release will not be soon?

I guess it depends on when this gets merged, whether more PRs show up while this is open / shortly after it is merged, whether I have time to do a release, ... :)

